PR #3, opened 2017-08-05
Add better solution for cross-platform font renderings

jiraiyame:
this feature will support in next generation css, ref: https://drafts.csswg.org/css-fonts-4/#system-ui-def

### DIFF
--- a/components/global/global.css
+++ b/components/global/global.css
@@ -33,7 +33,7 @@
 .font {
   color: text-color;
 
-  font-family: -apple-system, BlinkMacSystemFont, "Helvetica Neue", Arial, sans-serif;
+  font-family: system-ui, Arial, sans-serif;
 }
 
 /* To be used at large sizes */
--- a/components/global/global.scss
+++ b/components/global/global.scss
@@ -36,7 +36,7 @@ $ring-font-size-smaller: 11px;
 
 @mixin ring-font($font-size: $ring-font-size) {
   color: $ring-text-color;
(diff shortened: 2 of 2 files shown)


Hypnosphi:
In fact, it can be just `font-family: system-ui`, because we have cssNext in the project: http://cssnext.io/features/#system-ui-font-family

jiraiyame:
OK, i am missing that. One more thing, consider whether it is necessary to keep the font like `sans-serif` as a general fallback mechanism?

Hypnosphi:
I think it won't hurt to keep it

jiraiyame:
@Hypnosphi Changed done!

components/global/global.css
```diff
@@ -33,7 +33,7 @@
 .font {
   color: text-color;
 
-  font-family: -apple-system, BlinkMacSystemFont, "Helvetica Neue", Arial, sans-serif;
+  font-family: -apple-system, Arial, sans-serif;
```
Hypnosphi: I guess `-apple-system` is a typo =)

jiraiyame: wait a moment, i also find this mistake😅

jiraiyame:
@Hypnosphi this is the final change i guess 😆

Hypnosphi:
Thank you, looks great (in any sense)!

Leonya:
@jiraiyame unfortunately, we are going to revert this PR – the list of fonts we had before was a carefully considered one and we don't want any unexpected additions. For example, we have tested Segoe UI about a year ago and decided it's not a good fit for our UIs.

PR #4, opened 2017-08-15
Update firefox to 47.0 in gemini config

leonsabr:
It turned out that Sauce Labs provides Firefox 47.0 (Windows 10). Fortunately, there are no image diffs between 42 and 47.

Gemini does not support Firefox 48+ because `gemini` uses `wd` (https://github.com/admc/wd) package which does not support Selenium 3.0 yet. No estimates from gemini side.


leonsabr:
One failed test in IE11, not in firefox.

PR #1024, opened 2019-07-04
storybook: fix auth dialog dissapearing after swithcing to another section

edoroshenko:
Currently if you select auth service in storybook and then switch to another section, the auth dialog will remain.
Fixed this by calling remove callback in unmount hook.


Hypnosphi:
@huston007 I remember we had an idea to clean up everything on story change

huston007:
@Hypnosphi I believe we do, but there are some singleton services which can't be reset from storybook like `alertService`, that `auth-dialog-service` in this PR, maybe more. I don't se clear way to reset them without explicit calls, to be honest

PR #1445, opened 2019-11-14
fix - RG-2015 Jumping cursor in query assist

xel23:
fix - RG-2015 Jumping cursor in query assist


huston007:
@xel23 Thank you very much for preparing this fix!

PR #1544, opened 2019-12-20
Date Time Picker

subelta:
1. Fixed and validated keyboard input for Date Picker, 
2. Added the ability to input time (props withTime and time). If one tries to use it with range, only range functionality will work. Clearable works


.idea/cssdialects.xml
```diff
@@ -1,6 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
huston007: I believe even if this change is needed, it is out of scope of this pull request; May we revert this not related change?

components/date-picker/date-input.js
```diff
@@ -14,6 +14,7 @@ export default class DateInput extends Component {
     text: PropTypes.string,
     hoverDate: dateType,
     date: dateType,
+    time: PropTypes.string,
```
huston007: As far as I know, "date" prop is a JS/MomentJS Date, so it could hold both Date and Time inside. Maybe we don't have to split this single entity to "date" and "time"?

components/date-picker/date-picker.css
```diff
@@ -108,6 +108,25 @@
...
+}
+
+.timeInput {
+  position: relative;
+
+  &::before {
+    position: absolute;
+    left: calc(-1 * unit);
+
+    content: ',';
```
huston007: Always having comma looks weird when, for example, there is no date set
image

components/date-picker/date-picker.examples.js
```diff
@@ -49,6 +49,40 @@ singleDate.story = {
...
+        </div>
+      );
+    }
+  }
+
+  return <DatePickerExample/>;
+};
+
+singleDateAndTime.story = {
+  hermione: {skip: true},
```
huston007: Sorry, I gave you a wrong code sample: it should be `parameters: {hermione: {skip: true}}`

.idea/cssdialects.xml
subelta: Yeah, sure, I actually don't know how that happened

components/date-picker/date-picker.css
subelta: It was made in analogy to clearable range with only difference in placeholders presence

Screenshot 2019-12-24 at 14 09 42

But since mockup says range should behave like you suggested - I'll look into both

Screenshot 2019-12-24 at 14 12 10

components/date-picker/date-picker.examples.js
subelta: will fix

components/date-picker/date-input.js
subelta: I'll do the math about this one

.idea/cssdialects.xml
subelta: fixed

components/date-picker/date-input.js
subelta: The person I'm doing this component for would like to see date and time as different entities. 
Also i think
- having the ability to chose whether to specify time and date or date only gives flexibility and keeps things backwards compatible
- since with range=true usecase time is unnecessary and is not supposed to work, merging date and time would create inconsistency ({date} would in fact be date and time and {from} and {to} are just dates)

huston007: Okay, makes sense. Thank you for investigating this!

components/date-picker/date-picker.css
subelta: I removed comma (withTime) and dash (range) in some usecases, one of them is when both inputs are empty

Screenshot 2019-12-27 at 12 52 08

I decided to not delete the second input when props are empty as these usecases are handled well enough in both range and withTime, they also give flexibility of what to pick first

components/date-picker/date-picker.examples.js
subelta: done

subelta:
I think everything is ready. I reviewed all the possible usecases, introduced more consistent behaviour and fixes across them and wrote some simple tests

Regarding your suggestions answered in corresponding discussions, would appreciate the feedback

components/date-picker/date-picker.css
huston007: Looks much better, thanks!

subelta:
Ok, will look into it

huston007:
I've made a merge preview branch and am waiting for tests to pass. After that, we can merge it
https://teamcity.jetbrains.com/buildConfiguration/JetBrainsUi_RingUi_AllChecks/2718587?buildTab=overview

subelta:
Thanks!

PR #2324, opened 2020-09-15
Added renderEmpty={() => ''} prop

HelenOne:


components/table/table.examples.js
```diff
@@ -308,3 +308,43 @@ class MultiTableDemo extends Component {
 }
 export const multiTable = () => <MultiTableDemo/>;
 multiTable.storyName = 'multi table';
+
+
+class EmptyTableDemo extends Component {
+
+  state = {
+    selection1: new Selection({data: data1}),
+    selection2: new Selection({data: data2})
```
andrey-skl, line 317: Sorry, it isn't clear for me, why do we need 2 selections here?

components/table/table.examples.js
```diff
@@ -308,3 +308,43 @@ class MultiTableDemo extends Component {
...
+  };
+
+  columns2 = [
```
andrey-skl, line 320: Why "2"?) Looks like there are no "1"

andrey-skl (approved):

PR #2818, opened 2021-01-15
A render prop was introduced for date-popup component

dmitry-mashkov:
Sometimes it is necessary to involve specific components into the date popup. For example, buttons with relative periods (or dates) such as Today, Yesterday, Last Month, etc. We should not foresee all possible scenarios and visual interpretations, but at least we ought to give the possibility of displaying them.
I solve the issue using render props

### DIFF
--- a/components/date-picker/date-popup.js
+++ b/components/date-picker/date-popup.js
@@ -33,7 +33,7 @@ export default class DatePopup extends Component {
     time: PropTypes.string,
     from: dateType,
     to: dateType,
-    render: PropTypes.func,
+    renderAfterCalendar: PropTypes.func,
     displayFormat: PropTypes.func,
     parseDateInput: PropTypes.func,
     onChange: PropTypes.func,
@@ -394,7 +394,7 @@ export default class DatePopup extends Component {
           <Years {...calendarProps}/>
         </div>
 
-        {this.props.render && this.props.render(this.state)}
+        {this.props.renderAfterCalendar && this.props.renderAfterCalendar(this.state)}
       </div>
     );
   }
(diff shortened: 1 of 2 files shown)


components/date-picker/date-popup.js
```diff
@@ -33,6 +33,7 @@ export default class DatePopup extends Component {
     time: PropTypes.string,
     from: dateType,
     to: dateType,
+    render: PropTypes.func,
```
andrey-skl, line 36: I like the idea in general but think naming could be more specific to be friendly for further extensions: say tomorrow we want to add custom content above the picker and we would have to name it like "renderBeforePicker". Then we would have "render" and "renderBeforePicker" and this naming seems confusing.

Maybe we would better rename "render" prop to "renderAfterPicker"?

andrey-skl (changes requested):

components/date-picker/date-popup.js
dmitry-mashkov: Agree, it's better to change naming.
What do you think about "renderAfterCalendar" ? Because when we open the date popup we see the date input, then weekdays, then the calendar with a list of months and years on its right. And then goes the place where visually I put a custom component.
Do you think "picker" wouldn't be confused with the date input ?

andrey-skl: Sure, this name is even better. Thanks!

dmitry-mashkov: done

andrey-skl (approved):

Hypnosphi (approved):

PR #2851, opened 2021-01-28
i18n support was included for DatePicker

dmitry-mashkov:


components/date-picker/date-input.js
```diff
@@ -18,12 +18,22 @@ export default class DateInput extends React.PureComponent {
...
     hidden: PropTypes.bool,
+    translations: PropTypes.object,
     onInput: PropTypes.func,
     onActivate: PropTypes.func,
     onConfirm: PropTypes.func,
     onClear: PropTypes.func
   };
 
+  static defaultProps = {
+    translations: {
```
andrey-skl, line 29: It maybe makes sense to keep default values in the root component – just to see all wordings in one place. Don't you think?

dmitry-mashkov: I'm not sure. As long as we consider DateInput to be hypothetically used on its own it would make more sense to keep its default props by its side. Do we ?

PR #2884, opened 2021-02-03
RG-2135 Make tabs self adjustable to respect the available width

Blacknikomo:


andrey-skl (approved):

components/tabs/tab-autocollapse-titles.js
```diff
@@ -0,0 +1,541 @@
...
+    let resizeObserver = null;
+
+    if (measureRef.current === null) {
+      return;
+    }
+
+    resizeObserver = new ResizeObserver(entries => {
+      entries.forEach(entry => {
+        cancelAnimationFrame(animationFrame);
+        animationFrame = requestAnimationFrame(() => adjustTabs(entry));
```
Hypnosphi, line 470: this can be the source of the performance issue you mentioned. Please use fastdom for batching requestAnimationFrame call

components/tabs/tab-autocollapse-titles.js
```diff
@@ -0,0 +1,541 @@
...
+      });
+    });
+
+    if (typeof MutationObserver !== 'undefined') {
+      mutationObserver = new MutationObserver(() => {
+        cancelAnimationFrame(animationFrame);
```
Hypnosphi, line 476: Why is it needed? Shouldn’t ResizeObserver be enough

components/tabs/tab-autocollapse-titles.js
```diff
@@ -0,0 +1,541 @@
...
+  selected,
+  onSelect,
+  moreClassName,
+  moreActiveClassName,
+  morePopupClassName
+}) => {
+  const [elements, dispatch] = React.useReducer(visibilityReducer, DEFAULT_STATE);
+  const [preparedElements, setPreparedElements] = React.useState({visible: [], hidden: []});
+
+  const elementsToRender = useDebounce(preparedElements, DEFAULT_DEBOUNCE_INTERVAL);
```
Hypnosphi, line 356: I think it’s better to debounce the effect function that calls `setPreparedElements` (https://github.com/JetBrains/ring-ui/blob/ae126037246695f89c3e074e5b30d716502da0a0/components/tabs/tab-autocollapse-titles.js#L379-L406)

Hypnosphi (changes requested):
see comments

components/tabs/tab-autocollapse-titles.js
Blacknikomo: Initially it was supposed to detect changes in DOM, for example, in case the Custom Title changes it's content reflecting the 'selected' state change. But the truth is - this effect is used to adjust tabs, but not re-measure them, so I agree that it's redundant here. Thanks!

components/tabs/tab-autocollapse-titles.js
```diff
@@ -0,0 +1,511 @@
...
+    }
+    default:
+      throw new Error();
+  }
+}
+
+const useAdjustHandler = ({
+  measureRef, elements, children, selectedIndex, dispatch
+}) => React.useCallback(entry => {
+  const containerWidth = entry ? entry.contentRect.width : measureRef.current.offsetWidth;
```
Hypnosphi, line 273: after you removed MutationObserver, `entry` is always truthy, so the `measureRef.current.offsetWidth` part is not needed

components/tabs/tab-autocollapse-titles.js
```diff
@@ -0,0 +1,511 @@
...
+    let resizeObserver = null;
+
+    if (measureRef.current === null) {
+      return;
+    }
+
+    resizeObserver = new ResizeObserver(entries => {
+      entries.forEach(entry => {
+        fastdom.clear(measureTask);
+        measureTask = fastdom.measure(() => adjustTabs(entry));
```
Hypnosphi, line 453: `adjustTabs` uses `dispatch`, so it should be `mutate` not `measure`

Hypnosphi (changes requested):

components/tabs/tab-autocollapse-titles.js
Hypnosphi: Anyway, if content change affects element size, `ResizeObserver` will catch that

components/tabs/tab-autocollapse-titles.js
Blacknikomo: Should not it be like:
`fastdom.measure(() => {
       ....
       fastdom.mutate(() => {
           ....
       });
}`
       
Not simply 'fastdom.mutate'?

Hypnosphi: just `mutate` should be enough, I don't see any measuremens there

components/tabs/tab-autocollapse-titles.js
```diff
@@ -0,0 +1,512 @@
...
+
+import fastdom from 'fastdom';
+
+import Dropdown from '../dropdown/dropdown';
+
+import Link from '../link/link';
+
+import PopupMenu, {ListProps} from '../popup-menu/popup-menu';
+
+import Icon from '../icon';
```
Hypnosphi, line 15: this import is deprecated and removed in 4.0, please use `../icon/icon` instead

Hypnosphi (approved):

Hypnosphi:
The tests are failing because there is no ResizeObserver in JSDOM. You can add some kind of a mock to `setupFiles` in `jest.config.js` to fix it:

```js

class ResizeObserver {
    observe() {
        // do nothing
    }
    unobserve() {
        // do nothing
    }
    disconnect() {
       // do nothing
    }
}

window.ResizeObserver = ResizeObserver;
```

PR #3542, opened 2021-06-01
htmlType property was introduced to Button

dmitry-mashkov:


andrey-skl (approved):

Hypnosphi:
@dmitry-mashkov you can just use `type`, it would apply from `restProps`

dmitry-mashkov:
@Hypnosphi you are right, it would do. I am cancelling the PR then

PR #3807, opened 2021-08-17
RG-2161: fixed datepicker broken styles

zpix1:
Made some selectors for `DatePicker` component more specific, now it looks like in stable version. 

However, it doesn't fix the problem of broken order of selectors in `style.css` bundle, which appears because of using css modules and the way loader processes them. Maybe some other components could be affected.

### DIFF
--- a/components/date-picker/date-picker.css
+++ b/components/date-picker/date-picker.css
@@ -527,7 +527,7 @@
   line-height: yearHeight;
 }
 
-.currentYear {
+.currentYear.currentYear {
   cursor: auto;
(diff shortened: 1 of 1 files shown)


andrey-skl:
@zpix1 I'm afraid "selected year highlight" styling has been lost in this changes. You could see visual regression comparsion here: https://teamcity.jetbrains.com/buildConfiguration/JetBrainsUi_RingUi_GeminiTests/3583603?buildTab=report_project940_Visual_Regression

PR #4696, opened 2022-07-29
page size selector problem - [RG-2199]

lampmaster:
PageSizeSelector shouldn't disappear when the page size is equal to the "total" prop


Hypnosphi (approved):

lampmaster:
I changed the number of elements so visual regression tests didn't passed

PR #4902, opened 2022-10-07
Fix preventDefault in passive event

andreyvolfman:


andrey-skl:
@andreyvolfman I'm curious what is the context of this change : )

andrey-skl (approved):

PR #4996, opened 2022-11-09
RG-2221 fixes datepicker incorrect date selection

srsatt:


andrey-skl (approved):

andrey-skl:
@Hypnosphi sorry I didn't notice you are still reviewing it. If you still has concerns, please raise so we may fix it in master

PR #5034, opened 2022-11-23
Added error support for the input based select component

denbezrukov:


src/select/select.tsx
```diff
@@ -1247,47 +1250,61 @@ export default class Select<T = unknown> extends Component<SelectProps<T>, Selec
...
+                ? Object.keys({
+                  ...this.getShortcutsMap(),
+                  ...this._popup?.list?.shortcutsMap
+                })
+                : undefined}
+              afterInput={this.props.type === Type.INPUT && iconsNode}
             />
+            {this._renderPopup()}
+          </div>
+          {this.props.error && (
```
denbezrukov, line 1297: I used the same method as added error for this component
https://github.com/JetBrains/ring-ui/blob/c40ca5009296e983b7491f729daed01268460ba3/src/editable-heading/editable-heading.tsx#L168-L170

andrey-skl (approved):

PR #5527, opened 2023-04-26
Add I18n layer for Ring UI

andrey-skl:


andrey-skl:
I've made an draft of i18n implementation. Could you folks please check it out and let me know if it good in general?

src/header/profile.tsx
```diff
@@ -173,33 +182,33 @@ export default class Profile extends PureComponent<ProfileProps> {
     const items = [
       showApplyChangedUser && {
         rgItemType,
-        label: translations.applyChangedUser || 'Apply changed user',
+        label: translations?.applyChangedUser ?? messages.applyChangedUser,
```
vkrol, line 185: What do you think about using the same style everywhere?
```suggestion
        label: (translations ?? messages).applyChangedUser,
```

src/pager/pager.tsx
```diff
@@ -165,6 +164,7 @@ export default class Pager extends PureComponent<PagerProps> {
   }
 
   getPagerLinks() {
+    const messages = this.context as Messages;
```
vkrol, line 167: In order not to duplicate the type assertions we can do it the following way. This is the officially recommended way of typing the context in class components.
```ts
static contextType = I18nContext;
declare context: React.ContextType<typeof I18nContext>;

const messages = this.context;
```

vkrol: We can even do this:
```ts
static contextType = I18nContext;
declare context: React.ContextType<typeof Pager.contextType>;
```

Hypnosphi (approved):

src/user-card/card.tsx
```diff
@@ -33,11 +37,23 @@ export interface UserCardWording {
 }
 export interface UserCardProps extends HTMLAttributes<HTMLDivElement> {
   user: UserCardUser
-  wording: UserCardWording
+  wording?: UserCardWording | null | undefined // TODO: "wording" isn't the right name, to be replaced with "translations" in 6.0
```
vkrol, line 40: What about using JSDoc `@deprecated` tag in such cases?
```suggestion
  /** @deprecated will be replaced with `translations` in 6.0 */
  wording?: UserCardWording | null | undefined
```
image

src/user-card/card.tsx
```diff
@@ -56,39 +72,28 @@ export default class UserCard extends PureComponent<UserCardProps> {
...
 
   copyEmail = () => {
-    const wording = {...UserCard.defaultProps.wording, ...this.props.wording};
+    const messages = this.context as Messages;
+    const messageOverrides = this.props.wording || this.props.translations;
+
+    const wording = messageOverrides ?? messages;
     clipboard.copyText(
       this.props.user.email || '', wording.copiedToClipboard, wording.copingToClipboardError
     );
```
vkrol, line 88: We can already get rid of the term `wording` within the component:
```suggestion
    const messages = this.props.wording ?? this.props.translations ?? this.context;

    clipboard.copyText(
      this.props.user.email || '', messages.copiedToClipboard, messages.copingToClipboardError
    );
```
Once we have deleted the `wording` prop, it will be sufficient to simplify the expression for calculating the value of the `messages`:
```ts
    const messages = this.props.translations ?? this.context;
```

vkrol:
@andrey-skl Do I understand correctly that every time a new component is added that uses new localization messages, or when a new message is added to an existing component, it will be a breaking change for the project if it uses the new i18n functionality because new fields will be added to the `Messages` object that is passed to `setTranslations`? I remember that we discussed it at the meeting, but I don't remember what we came to.

src/header/profile.tsx
andrey-skl: Good point, thanks!

src/pager/pager.tsx
andrey-skl: Addressed in https://github.com/JetBrains/ring-ui/pull/5542

src/user-card/card.tsx
andrey-skl: If we delete "wording" prop, that would be breaking change, that's why we have to keep it for a while

src/header/profile.tsx
andrey-skl: Addressed in https://github.com/JetBrains/ring-ui/pull/5542

andrey-skl:
@andrey-skl Good question about breaking changes. I think we've addressed it in https://github.com/JetBrains/ring-ui/pull/5542

src/user-card/card.tsx
vkrol: I'm not removing the `wording` prop in my suggestion. Sorry for the confusion. 

> Once we have deleted the wording prop, it will be sufficient to simplify the expression for calculating the value of the messages:

It was about the future major version of the library. But we can now reduce the use of `wording` in this component. That was my suggestion.

andrey-skl: Got it!

I believe it is moreless OK now in a new PR 5542, how do you think?

https://github.com/JetBrains/ring-ui/blob/6d16d099f9ca352f5c88000e7a0677dfb4faed4e/src/user-card/card.tsx#L82

vkrol: In my opinion it's better to use something like this:
```ts
const translations = this.props.translations || this.props.wording; 
```
because `translations` is a canonical term in the codebase, but this is a very minor point, so let's leave it as that :)

PR #5542, opened 2023-05-02
RG-2276 update API of translation layer to meet requirements

andrey-skl:
On our weekly meeting, we've agreed to:

1. Change usages of `message.smth` to some function `getMessage('smth')`, because that adds more centralised control
2. In that `getMessage` function, check if we have that key overriden. If not, print warning
2. On type level, have regular "Optional" type `Messages`, so you can skip some messages, and a dedicated `MessagesStrict` so you can opt-in for strict translations typing
3. Add a dedicated Readme for i18n layer

### DIFF
--- a/src/auth/auth__core.ts
+++ b/src/auth/auth__core.ts
@@ -4,7 +4,7 @@ import HTTP, {HTTPAuth, RequestParams} from '../http/http';
 import promiseWithTimeout from '../global/promise-with-timeout';
 import AuthDialogService from '../auth-dialog-service/auth-dialog-service';
 
-import {getTranslations} from '../i18n/i18n';
+import {getTranslations, getTranslationsWithFallback, translate} from '../i18n/i18n';
 
 import AuthStorage, {AuthState} from './storage';
 import AuthResponseParser, {AuthError} from './response-parser';
@@ -68,7 +68,7 @@ export interface LoginFlowClass {
   new (
     requestBuilder: AuthRequestBuilder,
     storage: AuthStorage,
-    translations: AuthTranslations,
+    translations: AuthTranslations
   ): LoginFlow
 }
 
@@ -269,7 +269,7 @@ export default class Auth implements HTTPAuth {
       this._embeddedFlow = new this.config.EmbeddedLoginFlow(
         this._requestBuilder,
         this._storage,
-        this.config.translations ?? getTranslations()
+        this.config.translations ?? getTranslationsWithFallback()
       );
     }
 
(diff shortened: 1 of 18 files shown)


src/i18n/i18n.ts
```diff
@@ -1,8 +1,21 @@
...
-export type Messages = typeof defaultMessages;
+export type MessagesStrict = typeof defaultMessages;
 
-let messages = defaultMessages;
+export type Messages = Partial<MessagesStrict>;
+
+let messages: Messages = defaultMessages;
+
+function warnMissedKeyOnce(key: keyof MessagesStrict) {
+  const warned = new Set<string>();
```
vkrol, line 10: This variable must be placed at the topmost level of the module.

vkrol (changes requested):

src/i18n/README.md
```diff
@@ -1,3 +1,49 @@
...
+  login: 'Inloggen',
+  ...
+}
+
+setTranslations(localised);
+```
+
+3. [Optional] Add I18nContextHolder into your root tree. 
+   If you skip this step, messages from singleton service would be used.
+   Note: setTranslations() calls won't cause React tree rerender, so you have to call it before first render.
```
vkrol, line 30: I don't think it would work. If an user doesn't render `I18nContextHolder`, the default value of `messages` in the context will never be overwritten by a `setTranslations` call, because the context will be initialized when the `i18n-context.ts` module is imported and we cannot call `setTranslations` before that. I created a minimal reproducible example https://codesandbox.io/s/sad-field-c2p9p0.

src/i18n/README.md
```diff
@@ -1,3 +1,46 @@
...
+const localised = {
+  login: t`Log in`,
+  ...
+}
+```
+
+2. Pass it to i18n singleton, using `setTranslations`. That would provide localisation to non-react components, like Auth.
+
+```
+// Your messages object should have loalised strings at that moment
```
vkrol, line 19:
```suggestion
// Your messages object should have localised strings at that moment
```

vkrol (changes requested):

src/auth/auth__core.ts
vkrol, line 1: What is the reason for the changes in this file? Is it unrelated to https://github.com/JetBrains/ring-ui/pull/5542#discussion_r1182546677?

vkrol (changes requested):

src/auth/auth__core.ts
vkrol: Never mind. I read the commit message and now I understand :)

vkrol (approved):

PR #5808, opened 2023-06-16
RG-2289 replace @nest with standardized nesting syntax

vkrol:


stylelint.config.js
```diff
@@ -1,6 +1,9 @@
 module.exports = {
   extends: '@jetbrains/stylelint-config',
+  plugins: ['@csstools/stylelint-no-at-nest-rule'],
   rules: {
+    '@csstools/stylelint-no-at-nest-rule': true,
```
vkrol, line 5: It's up for discussion :)

src/heading/heading.css
```diff
@@ -9,30 +9,30 @@
     margin-top: 0;
   }
 
-  @nest h1&, h2& {
+  &:is(h1, h2) {
```
vkrol, line 12: Note that the native syntax does not allow us to write it like this:
```css
.heading {
  &h1 {}
}
```

stylelint.config.js
andrey-skl: Since it protects us from unwanted syntax usage, I don't see reason against that rule : )

andrey-skl (approved):

src/tags-input/tags-input.css
```diff
@@ -61,7 +61,7 @@
 
   caret-color: var(--ring-main-color);
 
-  @nest [dir=rtl] & {
+  [dir="rtl"] & {
```
vkrol, line 64: Oops, it should be:
```suggestion
  [dir=rtl] & {
```
I will fix it.

PR #5944, opened 2023-07-10
RG-2291 implement primary type of controls labels

andrey-skl:
See https://youtrack.jetbrains.com/issue/RG-2291/Form-variant-for-text-fields-select-fields


src/control-label/control-label.css
```diff
@@ -0,0 +1,21 @@
+.label {
+  display: block;
+
+  margin-bottom: calc(var(--ring-unit) * 0.5);
+
+  color: var(--ring-secondary-color);
+
+  font-size: var(--ring-font-size-smaller);
+  line-height: var(--ring-line-height-lowest);
+}
```
vkrol, line 10: I think it's better to put the secondary type styles in a separate CSS class to make it easier to understand which final set of styles will be applied to the primary type labels.
```css
.label {
  display: block;

  margin-bottom: calc(var(--ring-unit) * 0.5);
}

.primaryLabel {
  color: var(--ring-text-color);

  font-size: var(--ring-font-size);
  line-height: var(--ring-line-height);
}

.secondaryLabel {
  color: var(--ring-secondary-color);

  font-size: var(--ring-font-size-smaller);
  line-height: var(--ring-line-height-lowest);
}

.disabledLabel {
  color: var(--ring-disabled-color);
}
```
```tsx
const classNameByType: Record<LabelType, string> = {
  [LabelType.SECONDARY]: styles.secondaryLabel,
  [LabelType.PRIMARY]: styles.primaryLabel
};

export const ControlLabel: React.FC<ControlLabelProps> = (
  {children, type = LabelType.SECONDARY, disabled, ...rest}
) => (
  <label
    className={classNames(styles.label, classNameByType[type], {
      [styles.disabledLabel]: disabled
    })}
    {...rest}
  >{children}</label>
);
```
What do you think about that?

vkrol (changes requested):

vkrol (approved):

vkrol (approved):

PR #6270, opened 2023-09-18
RG-2315 Do not keep DatePicker popup mounted

vkrol:
The `hidden` prop will now always be `false`, so I deleted it.


andrey-skl:
> The `hidden` prop will now always be `false`, so I deleted it.

That is a good point, but I am afraid it may be concidered as breaking change to remove prop (even though it does nothing). What do you think, can it break something?

vkrol:
I think we only consider this a breaking change if we consider the `DatePopup` component as part of the public API.

andrey-skl:
@vkrol Got it! That is a good point, I think DatePopup should be concidered internal component and we should not concidered it Props public, so the change is not breaking

andrey-skl (approved):

PR #6496, opened 2023-10-30
Add wrapper for List element to create custom Tooltip for example

Xatabch:
Add wrapper for List element to create custom Tooltip for example. Task: https://youtrack.jetbrains.com/issue/JT-77244/Subsystem-name-is-not-fully-visible-in-new-UI


src/list/consts.ts
```diff
@@ -64,6 +66,7 @@ export type ListDataItem<T = unknown> =
   originalModel?: never
   LinkComponent?: ComponentType<ClickableLinkProps> | string | null | undefined
   template?: ReactNode | ((props: ListDataItemProps<T>) => ReactNode)
+  wrapper?: (children: ReactNode) => ReactElement
```
andrey-skl, line 69: I think, it may return ReactNode too, why not
```
(children: ReactNode) => ReactNode
```

PR #6525, opened 2023-11-04
Add multiline input for editable heading. Task JT-77542

Xatabch:


src/editable-heading/editable-heading.tsx
```diff
@@ -34,6 +36,8 @@ export type EditableHeadingProps = Omit<InputHTMLAttributes<HTMLInputElement>, '
   'data-test'?: string | null;
   error?: string;
   multiline?: boolean;
+  multilineInput?: boolean;
```
andrey-skl, line 39: What would you think of using single `multiline` prop for that purpose? Otherwise it may be weird to have `multiline={false} multilineInput={true}`, isn't it?

src/editable-heading/editable-heading.tsx
```diff
@@ -121,15 +130,33 @@ export const EditableHeading = (props: EditableHeadingProps) => {
...
+
+  const onInputBlur = React.useCallback(
+    (e: React.FocusEvent<HTMLInputElement | HTMLTextAreaElement>
+    ) => {
+      setIsInFocus(false);
+      onBlur?.(e);
+    }, [onBlur]);
+
+  textAreaRef?.current?.addEventListener('scroll', () => {
+    const scrollHeight = textAreaRef?.current?.scrollHeight || 0;
```
andrey-skl, line 148: I believe, first optional chaining is redundant (`textAreaRef?.`)

also, wouldn't it be simplier to just have 1 check in the beginning?|
```
if (!textAreaRef.current) {
  return
}
```

src/editable-heading/editable-heading.tsx
```diff
@@ -121,15 +130,33 @@ export const EditableHeading = (props: EditableHeadingProps) => {
...
+    const clientHeight = textAreaRef?.current?.clientHeight || 0;
+    const scrollTop = textAreaRef?.current?.scrollTop || 0;
+
+    const isScrolledToBottom = scrollHeight - clientHeight <= scrollTop;
+
+    if (isScrolledToBottom) {
```
andrey-skl, line 154: I can't understand why we have to remove `textareaWrapper` class when scrolled to bottom, maybe you can help me to understand?

src/editable-heading/editable-heading.tsx
```diff
@@ -153,16 +180,35 @@ export const EditableHeading = (props: EditableHeadingProps) => {
...
+                    autoFocus={autoFocus}
+                    data-test={dataTest}
+                    disabled={isSaving}
+                    {...restProps}
+                    onFocus={onInputFocus}
+                    onBlur={onInputBlur}
+                  />
+                )
+                : (
+                  <div ref={textAreaWrapperRef} className={styles.textareaWrapper}>
```
andrey-skl, line 197: Is the `div` wrapper absolutely required here? If you maybe can share what is it's purpose, we can understand it better

andrey-skl:
Corresponding task in Ring UI https://youtrack.jetbrains.com/issue/RG-2322/EditableHeader-should-support-multiline-editing

src/editable-heading/editable-heading.tsx
Xatabch: Yes I agree with you, I wanted to keep backwards compatibility so that if someone was already using multiline separately, it wouldn't be a surprise to them that it's now also multiline editing. But in general, I can use it as a single flag

src/editable-heading/editable-heading.tsx
Xatabch: Yes, it's for fading in from the bottom of the textarea
Screenshot 2023-11-04 at 18 59 04

src/editable-heading/editable-heading.tsx
Xatabch: Yes, the answer is in the previous comment

src/editable-heading/editable-heading.tsx
Xatabch: Fade should only exist when we're not at the bottom of our scroll

Xatabch: Here is the video to understand it better

Uploading Screen Recording 2023-11-04 at 19.06.16.mov…

Xatabch: Here is the video to understand it better

Uploading Screen Recording 2023-11-04 at 19.06.16.mov…

src/editable-heading/editable-heading.tsx
Xatabch: Unfortunately we can't apply that rule here. Because in my case we just won't hover lisner if we don't have ref on textarea. In your case, we will do a full return and won't render anything if we don't have a ref, which will break the component. Attached screenshot
Screenshot 2023-11-04 at 19 12 32

src/editable-heading/editable-heading.tsx
andrey-skl: Okay, thank you! Maybe we could add more context to that classname by naming it like `...withShadow`, but that's does not matter too much

src/editable-heading/editable-heading.tsx
andrey-skl: I understand the reasoning, thank you.

But in this specific case we may say "having no multilines in Input mode is a kind of bug", because I believe it is the expected behavior to have it. So I'd vote to have a single prop

src/editable-heading/editable-heading.tsx
```diff
@@ -121,15 +130,33 @@ export const EditableHeading = (props: EditableHeadingProps) => {
...
+    }, [onFocus]);
+
+  const onInputBlur = React.useCallback(
+    (e: React.FocusEvent<HTMLInputElement | HTMLTextAreaElement>
+    ) => {
+      setIsInFocus(false);
+      onBlur?.(e);
+    }, [onBlur]);
+
+  textAreaRef?.current?.addEventListener('scroll', () => {
```
andrey-skl, line 147: I think attaching listener every render (and never removing them) isn't really a good idea, because we will end up with dozens of listeners that are trying to do the same action. Can't we put this into a useEffect?

```
  useEffect(() => {
    const onScroll = (event) => console.info("scrolling", event);
      
    window.addEventListener('scroll', onScroll);
    
    return () => {
      window.removeEventListener('scroll', onScroll);
    }
  }, []);
  ```

andrey-skl (changes requested):

src/editable-heading/editable-heading.tsx
Xatabch: Oh, yes, I am very sorry. I forgot about that, thank you

Xatabch: fixed that

src/editable-heading/editable-heading.tsx
Xatabch: Is this comment should resolve or maybe I have something missed?

andrey-skl: Yep, thanks

src/editable-heading/editable-heading.tsx
andrey-skl: I think this is the last point left to address

Xatabch: done

andrey-skl (approved):

andrey-skl:
@Xatabch Thanks, greatly done!

There is just 1 visual test failing https://artifacts.teamcity.jetbrains.com/presignedTokenAuth/8722a632-1275-445d-9508-43a62beb675e/repository/download/JetBrainsUi_RingUi_GeminiTests/4373846:id/hermione-images/Components/Editable%20Heading%20Editable%20Heading/chrome/editable%20heading.diff.png

PR #6552, opened 2023-11-08
Make textarea stretched and some refactoring. JT-77542

Xatabch:


andrey-skl:
@Xatabch Do I understand it right the purpose of this PR is to make textarea auto-stretched while you're typing text?

Xatabch:
> @Xatabch Do I understand it right the purpose of this PR is to make textarea auto-stretched while you're typing text?

Yes and also add max height

andrey-skl (approved):

vkrol:
I think this kind of API would be hard to use. This is because in this prop you have to pass a magic number calculated depending on `line-height` (therefore, in order for the client of the component to know what prop value to pass to it, it has to examine the source code of the component), and when you change `line-height`, it has to be changed too. And the worst thing is that changing `line-height` inside the `EditableHeading` component will now become a breaking change, because of this you will have to necessarily change the `maxInputHeight` value for all clients.

What if we try to replace this prop with something like `maxRows`, and in the component itself calculate `max-height` based on this prop? We can use the new `lh` unit of measure (https://css-tricks.com/lh-and-rlh-units/) if the set of browsers we support allows it.

Xatabch:
> I think this kind of API would be hard to use. This is because in this prop you have to pass a magic number calculated depending on `line-height` (therefore, in order for the client of the component to know what prop value to pass to it, it has to examine the source code of the component), and when you change `line-height`, it has to be changed too. And the worst thing is that changing `line-height` inside the `EditableHeading` component will now become a breaking change, because of this you will have to necessarily change the `maxInputHeight` value for all clients.
> 
> What if we try to replace this prop with something like `maxRows`, and in the component itself calculate `max-height` based on this prop? We can use the new `lh` unit of measure (https://css-tricks.com/lh-and-rlh-units/) if the set of browsers we support allows it.

1. If you depend on something to calculate height, it may cause a breaking change. So, for example, you want to rely on line-height, but you don't take into account such a thing as font-size or the same kerning, which can also change and change the height of the line in some way. A task of this kind entails quite a lot of logic, which may not cover all the cases
2. At this point, if you at some point somewhere decide to change the line-height, for multiple, then if you do the task, you see that it visually changes, so if you change the line-height and it requires a change in height, you're more likely to correct that within the task
3. You have more flexible change control that's up to you, not the logic within

andrey-skl:
Dear folks, I would like to invite you to discuss the matter on next Ring UI weekly meeting

andrey-skl:
So we had a discussion on today's weekly and agreed to have `multilineInputRows` property back. And we will use `lh` units for implementation

PR #6559, opened 2023-11-09
RG-2329 Get rid of custom sniffr typings

vkrol:
We can remove our custom timings after upgrading to `sniffr@1.3.0` because the package now contains them https://unpkg.com/browse/sniffr@1.3.1/dist/sniffr.d.ts.


andrey-skl (approved):

vkrol:
@andrey-skl I believe this will not be a breaking change, because as I understand it, the library client could not import these enums and they do not exist in runtime, and the Sniffer class type is compatible with the new type from the library's own typings.

PR #6562, opened 2023-11-09
Make Welcome stories default

vkrol:
Screenshot 2023-11-09 at 20 07 41


andrey-skl:
Perfect catch, thank you

PR #6592, opened 2023-11-15
Made editable heading dependent on maxInputRows

Xatabch:


src/editable-heading/editable-heading.tsx
```diff
@@ -238,7 +238,7 @@ export const EditableHeading = (props: EditableHeadingProps) => {
                       onFocus={onInputFocus}
                       onBlur={onInputBlur}
                       onScroll={onInputScroll}
-                      style={{maxHeight: maxInputHeight}}
+                      style={{maxHeight: `calc(${maxInputRows} * 1lh)`}}
```
andrey-skl, line 241: Can't it be even simplier `${maxInputRows}lh`?

src/editable-heading/editable-heading.tsx
```diff
@@ -51,7 +51,7 @@ export const EditableHeading = (props: EditableHeadingProps) => {
     autoFocus = true, 'data-test': dataTest, error, disabled, multiline = false,
     renderMenu = () => null,
     onFocus, onBlur, onChange,
-    onScroll, maxInputHeight,
+    onScroll, maxInputRows,
```
andrey-skl, line 54: Don't we need to have some default here?

src/editable-heading/editable-heading.tsx
Xatabch: of corse

src/editable-heading/editable-heading.tsx
Xatabch: If I remember correctly, we decided to make the textarea have no constraints in case lh doesn't work, but I think the same behavior should also work for if we passed nothing in

andrey-skl: I agree on that. But is it working now? Looks like, if we pass nothing, it would be `max-height: undefinedlh;` which looks weird even if works fine

Xatabch: fixed that

andrey-skl: Thanks!

andrey-skl (approved):

src/editable-heading/editable-heading.stories.tsx
```diff
@@ -106,7 +106,7 @@ export const basic = () => {
       </ExampleEditableHeading>
       {lorem}
 
-      <ExampleEditableHeading multiline maxInputHeight={150}>
+      <ExampleEditableHeading multiline>
```
vkrol, line 109: I think it's a good idea to add the `maxInputRows` here or to make a new story in which that attribute will be passed.

vkrol (approved):

src/editable-heading/editable-heading.stories.tsx
Xatabch: done

vkrol: Thanks!

PR #6594, opened 2023-11-16
Fix scroll enter when not overflow. JT-78232

Xatabch:


src/editable-heading/editable-heading.tsx
```diff
@@ -102,6 +103,7 @@ export const EditableHeading = (props: EditableHeadingProps) => {
     'ring-js-shortcuts',
     styles.input,
     styles.textarea,
+    {[styles.textareaNotOverflow]: !isOverflow},
```
andrey-skl, line 106: I am not sure I understand why we need to disable scrollbar when text size is bigger than scrolling area. Could you please help me to understand?

Xatabch: Of course, in the task I pointed out, the problem is that in the case when we have a scrollbar but it is not shown, problems arise(you can see this in the task, Venya described it in detail there). With this approach this problem is solved in a classic input where textarea is used, only there is no height limit and so we don't need this logic there, but we can set height limits.

vkrol (approved):

PR #6864, opened 2024-01-23
<Collapse /> component

SmirnovW:
## Description
The Collapse component has been created.

## Motivation
This component is a crucial UI element for building interfaces and was missing in the RingUI. To implement the Collapse component, the Compound Component Pattern was used.

## Implementation
The Collapse component consists of one main component and two sub-components: 
1. **\<Collapse/\>** - This component provides the context/state to other components.
2. **\<CollapseControl/\>** - This component wraps any controller/trigger that triggers the collapse and expand functionality.
3. **\<CollapseContent/\>** - This component contains the collapsed content. 

All these components are sharing states using the React Context.

### Basic usage
```javascript
<Collapse>
 <CollapseControl>
  <Button>Show</Button>
 </CollapseControl>
 <CollapseContent>{text}</CollapseContent>
</Collapse>
```

### Collapse
| Attribute | Type                         | Description                                                                      | Default |
|-----------|------------------------------|----------------------------------------------------------------------------------|---------|
| children  | ReactNode                    | Should contain the CollspaseContent and CollapseControl components.              | -       |
| className | string                       | Wrapper's class name.                                                            | -       |
| onChange  | (collapsed: boolean) => void | This callback function will be called when the content is collapsed or expanded. | -       |
| duration  | number                       | Animation duration.                                                              | 200     |

### CollapseContent
 Attribute        | Type                                             | Description                     | Default 
------------------|--------------------------------------------------|---------------------------------|---------
 children         | ReactNode \| \(collapsed: boolean\) => ReactNode | The collapsed content\.         | \-      
 data\-test       | string \| null \| undefined                      | Data test attribute\.           | \-      
 minHeight        | number                                           | Min default height\.            | 0       
 disableAnimation | boolean                                          | Disables sollapsing animation\. | false   
 className        | string                                           | Wrapper's class name\.          | \-      

### CollapseConttrol
 Attribute  | Type                                             | Description                                     | Default 
------------|--------------------------------------------------|-------------------------------------------------|---------
 children   | ReactNode \| \(collapsed: boolean\) => ReactNode | The control will trigger collapsing/expanding\. | \-      
 data\-test | string \| null \| undefined                      | Data test attribute\.                           | \-      

### Basic
Screenshot 2024-01-23 at 18 39 47

### With default min height.
Screenshot 2024-01-23 at 18 40 03

### The trigger is under the content
Screenshot 2024-01-23 at 18 40 27


src/collapse/collapse-content.tsx
```diff
@@ -0,0 +1,127 @@
...
+    } else {
+      setShowFade(true);
+    }
+    onChange(collapsed);
+  }, [collapsed, onChange]);
+
+  useEffect(() => {
+    const ref = containerRef.current;
+    if (ref) {
+      ref.addEventListener('transitionstart', onTransitionStart);
```
andrey-skl, line 66: Wouldn't this event be triggered by bubbled event if some user content has been animated inside <CollapseContent/>?

src/collapse/collapse.stories.tsx
```diff
@@ -0,0 +1,177 @@
...
+  }
+};
+const text = `This is very long text! This is very long text! This is very long text! This is very
+            long text! This is very long text! This is very long text! This is very long text! This
+            is very long text! This is very long text! This is very long text! This is very long
+            text! This is very long text! This is very long text! This is very long text! This is
+            very long text! This is very long text! This is very long text! This is very long text!
+            This is very long text! This is very long text! This is very long text!`;
+
+export const Basic = () => (
```
andrey-skl, line 38: There is no screenshot gathered for this components, would you capture them?

src/collapse/utils.ts
```diff
@@ -0,0 +1,9 @@
+export function getElementHeight(element: HTMLElement | null) {
```
andrey-skl, line 1: You can maybe use this helper instead https://github.com/JetBrains/ring-ui/blob/6d38612f9deb7aa99a7a2e5b135fad920df8fdb4/src/global/dom.ts#L29

```
   getRect(element).height
```

src/collapse/collapse-content.tsx
```diff
@@ -0,0 +1,127 @@
...
+const DURATION_FACTOR = 0.5;
+const DEFAULT_HEIGHT = 0;
+const VISIBLE = 1;
+const HIDDEN = 0;
+
+type Props = {
+  minHeight?: number;
+  className?: string;
+  'data-test'?: string | null | undefined;
+  disableAnimation?: boolean;
```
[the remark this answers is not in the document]
SmirnovW: @andrey-skl, Thank you for bringing this up. I was also uncertain about the decision and was considering passing these props to the same component. 
I think it would be better to pass these two props to the main <Collapse/> component. 

As for having a single prop, I'm not sure, I believe it would be better to have two different components.
I think it would make the API more obvious. WDYT?

src/collapse/collapse-content.tsx
SmirnovW: @andrey-skl  Yes, you're right. To be honest, this addEventListener is redundant and can be removed.

src/collapse/collapse.stories.tsx
SmirnovW: @andrey-skl I'll add them, thank you 👍

src/collapse/utils.ts
SmirnovW: @andrey-skl Ah, I didn't know about this util. Yes, it can replace my custom getElementHeight util. Thank you!

src/collapse/collapse-content.tsx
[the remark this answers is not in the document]
andrey-skl: Two props for the main component would work I believe. Thanks!

src/collapse/collapse-control.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  const child = useMemo<React.ReactElement>(() => {
+    if (typeof children === 'function') {
+      return <p data-test={dataTests(COLLAPSE_CONTROL_TEST_ID, dataTest)}>{children(collapsed)}</p>;
+    }
+
+    return <p data-test={dataTests(COLLAPSE_CONTROL_TEST_ID, dataTest)}>{children}</p>;
+  }, [children, collapsed, dataTest]);
+
+  return cloneElement(child, {
```
skoch13, line 30: The current RingUI components do rely heavily on `cloneElement`, but it's worth noting that this API is [considered](https://react.dev/reference/react/cloneElement) legacy and fragile. For that reason, I would recommend [exploring](https://react.dev/reference/react/cloneElement#alternatives) the suggested alternatives when creating a new component.

skoch13 (commented):
Thank you for the PR ❤️

andrey-skl (approved):

vkrol (approved):

andrey-skl:
The PR seems to be fine in general, so let's have it merged. If more comments or concerns arise, welcome to post it here and we may address them in `master`

src/collapse/collapse-control.tsx
SmirnovW: @skoch13 Thank you for mentioning this. Sorry for the delay. I'll create another pull request to suggest improvements for this particular usage of cloneElement.

PR #7054, opened 2024-02-29
Add Slider component

kotofeyskaya:
## Description
'Slider' is a component that allows users to select a value from a range by dragging a 'thumb' along the track.

### Usage
`
 <Slider defaultValue={25} step={25} marks showTicks showTag/>
`

image
image
image

## API

Property | Type | Description | Default
-- | -- | -- | --
defaultValue | number \| number[] | The default value | -
value | number \| number[] | Value, that is used when the component is controlled | -
min | number | Min value of the slider | 0
max | number | Max value of the slider | 100
step | number | The interval between each value on the slider | 1
disabled | boolean | Disables user interaction | false
marks | boolean \| {value: number, label: ReactNode}[]  | If set to 'true', mark values are shown according to the 'step'. If set to an array, then the provided values are shown. Mark values are not shown if this prop is empty. | -
showTicks | boolean | Shows the tick marks on the track. Ticks  | false
showTag | boolean | Shows the tag of the selected value | false
className | string | Slider's class name. | -
renderTag | (value: number) => ReactNode | Used to modify tag's content | -
onChange | (value: number \| number[]) => void | Callback function triggered when the value changes | -


src/slider/slider.css
```diff
@@ -0,0 +1,176 @@
+@value dark from "../global/variables_dark.css";
+
+:root {
+  --ring-slider-thumb-color: var(--ring-content-background-color);
```
andrey-skl, line 4: How do you think, doesn't it make sense to define these variables not on the `:root` but on Slider container `.slider` instead? That way, it would be more local and easier to debug

src/slider/slider.tsx
```diff
@@ -0,0 +1,285 @@
...
+  onChange
+}) => {
+  const ref = useRef<HTMLDivElement>(null);
+  const previouslyDragged = useRef(false);
+  const [values, setValues] = useState(defaultValue ?? min);
+  const validValues: number[] = useMemo(
+    () => toRange(value ?? values, min, max),
+    [max, min, value, values]
+  );
+  const validStep = useMemo(() => (step < 0 ? 0 : step), [step]);
```
andrey-skl, line 68: I would maybe recomment avoid overusing `useMemo` for simple calculations, because it it makes no sense to remember things that are calculated in a single operation

src/slider/slider.tsx
```diff
@@ -0,0 +1,285 @@
...
+  const handleMouseMove = useCallback(({pageX}: MouseEvent) => {
+    const nextValues = adjustValues(validValues, ref, draggedIndex, pageX, max, min, validStep);
+    if (nextValues[0] > nextValues[1]) {
+      nextValues.reverse();
+      setDraggedIndex(prevState => (prevState === 0 ? 1 : 0));
+    }
+    handleValueChange(nextValues);
+  }, [validValues, draggedIndex, max, min, validStep, handleValueChange]);
+
+  const handleKeyDown = useCallback(({key, currentTarget}: KeyboardEvent) => {
```
andrey-skl, line 144: I would request using `Shortcuts` [component](https://github.com/JetBrains/ring-ui/blob/master/src/shortcuts/shortcuts.ts) instead, because we try to implement shortcuts in a single way to make it unified and controlled

src/slider/slider.tsx
```diff
@@ -0,0 +1,285 @@
...
+        className={classNames(styles.track, {
+          [styles.rounded]: !showTicks,
+          [styles.disabled]: disabled
+        })}
+      />
+      {validValues.map((numValue, index) => {
+        const percent = toPercent(numValue, min, max);
+        return (
+          // eslint-disable-next-line react/no-array-index-key
+          <Fragment key={index}>
```
andrey-skl, line 224: Maybe `numValue` is better option for key?

kotofeyskaya: Yes, I thought about it, too, but it creates an ambiguous situation for a 'range' slider when we have two thumbs pointing at the same value and end up having two duplicate keys.

src/slider/slider.tsx
kotofeyskaya: Thank you for mentioning that, replaced it with Shortcuts

src/slider/slider.tsx
kotofeyskaya: You are right, thank you, fixed

src/slider/slider.css
kotofeyskaya: Good idea, this does make sense, fixed!

andrey-skl (approved):

PR #7163, opened 2024-03-25
encodeURL: do not encode $ dollar symbol because it is OK to be in query param

vkrol:


src/global/url.ts
```diff
@@ -128,8 +128,9 @@ export function parseQueryString(queryString: string | null | undefined) {
 }
 
 function customEncodeURIComponent(str: unknown) {
-  const commaRE = /%2C/g;
-  return encodeURIComponent(String(str)).replace(commaRE, ',');
+  return encodeURIComponent(String(str)).
+    replace(/%2C/g, ',').
```
vkrol, line 132: I think that the second parameter is enough to make it clear what is being replaced, so I got rid of the commaRE variable.

andrey-skl (approved):

PR #7319, opened 2024-04-30
feat(select): add overscrollPopupBehaviourNone prop

skoch13:


andrey-skl (approved):

src/select/select__popup.tsx
```diff
@@ -703,7 +707,8 @@ export default class SelectPopup<T = unknown> extends PureComponent<SelectPopupP
   tags: PropTypes.object,
   toolbar: PropTypes.node,
   topbar: PropTypes.node,
-  top: PropTypes.number
+  top: PropTypes.number,
+  overscrollListBehaviorNone: PropTypes.bool
```
vkrol, line 711: How can I, as a library user, pass this prop?

vkrol (changes requested):

src/select/select__popup.tsx
vkrol, line 711: I suggest naming this prop in a less technical way, such as `preventListOverscroll`.

vkrol (changes requested):

src/select/select__popup.tsx
skoch13: thanks, fixed

src/select/select.tsx
```diff
@@ -204,7 +204,8 @@ export interface BaseSelectProps<T = unknown> {
   top?: number | undefined
   left?: number | undefined
   renderOptimization?: boolean | undefined
-  ringPopupTarget?: string | null | undefined
+  ringPopupTarget?: string | null | undefined,
+  overscrollPopupBehaviorNone?: boolean | undefined,
```
vkrol: The commas are redundant.

vkrol (changes requested):

vkrol (approved):

PR #7421, opened 2024-06-10
Remove Input .active CSS class usage

vkrol:
It was removed here https://github.com/JetBrains/ring-ui/commit/a45730a8db247a50406f09627c7e6489fe055a7f#diff-9571a6a46f39072419647c7b25d8afcfee0c28f631246b5e8e2bebdb974db8c4L160.

Technically, this is a breaking change, but I suspect that the `active` prop hasn’t been used for a long time. But we can keep the prop for now and make it optional if we’re unsure about it. What do you think?


andrey-skl:
I think I am okay with dropping it, we can say we fixed a bug that it wasn't dropped, let's just do it : )

vkrol:
@andrey-skl OK! Thanks.

PR #7475, opened 2024-06-28
Use stable composed ref in all components

vkrol:
This is the follow-up to https://github.com/JetBrains/ring-ui/pull/7458. I am doing the same for all components.

### DIFF
--- a/src/checkbox/checkbox.tsx
+++ b/src/checkbox/checkbox.tsx
@@ -6,7 +6,7 @@ import minusIcon from '@jetbrains/icons/remove-12px';
 
 import Icon from '../icon/icon';
 import {refObject} from '../global/prop-types';
-import composeRefs from '../global/composeRefs';
+import {createComposedRef} from '../global/composeRefs';
 
 import ControlHelp from '../control-help/control-help';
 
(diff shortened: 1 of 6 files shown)


src/global/composeRefs.ts
```diff
@@ -1,7 +1,7 @@
 import {Ref, MutableRefObject} from 'react';
 import memoizeOne from 'memoize-one';
 
-export default function composeRefs<T>(...refs: (Ref<T> | undefined)[]) {
+function composeRefs<T>(...refs: (Ref<T> | undefined)[]) {
```
Hypnosphi, line 4: please leave this export as it may be used outside of Ring UI

vkrol: Sure, this is a valid concern.

Although I would like to propose to forbid using of this function in the ring-ui repository via ESLint, because it's easy too easy to make a mistake and use it instead of createComposedRef in render function. What do you think about it? It was the main reason why I decided to remove the export of it.

Hypnosphi: You can deprecate it using `util-deprecate` and mark for removal in the next major version

Hypnosphi (approved):

src/global/composeRefs.ts
vkrol: Thanks, I will do it in a separate PR.

PR #7533, opened 2024-07-15
Color update

MStanzl:
Updated the colors to be in line with the IntUI color system.

[RG-2402](https://youtrack.jetbrains.com/issue/RG-2402) Adjust colors to be in line with the IntUI Color system


andrey-skl:
I've overcome this by patching sources locally:
Screenshot 2024-07-15 at 15 57 08

But not clear for me why it happens locally but not on CI. Maybe related to https://github.com/facebook/react/issues/24399

I'll get back to this if issue repeated

Hypnosphi:
> Maybe related to https://github.com/facebook/react/issues/24399

Yes, it's most likely related. I think we should just skip calling the methods if the dialog is already open

andrey-skl (approved):

Hypnosphi:
Reopened as #7548

PR #7550, opened 2024-07-22
feat(simple-table): add simple table component

skoch13:
https://youtrack.jetbrains.com/issue/RG-2457


andrey-skl (approved):

skoch13:
@andrey-skl @Hypnosphi it was decided to refactor the existing and new component once again:
* `wideFirstColumn` prop was added and set to `true` for compatibility
* `wideFirstColumn = false` is used for SimpleTable by default

PR #7573, opened 2024-07-25
TW-88703 Header: add vertical variant

Hypnosphi:
Includes changes from #7565, so that one should be merged first


andrey-skl:
> Includes changes from https://github.com/JetBrains/ring-ui/pull/7565, so that one should be merged first

Sure, I believe we will merge today or on Monday

Hypnosphi:
Included in #7578

PR #7614, opened 2024-08-05
RG-2341 put styles to layers: "ring-ui_global" for global variables and "ring-ui" for rest library styles

andrey-skl:
See https://youtrack.jetbrains.com/issue/RG-2341/Research-if-CSS-Cascade-Layers-may-help-solve-issues-with-style-order-and-style-overriding for details

See illustration on how it looks:
Screenshot 2024-08-05 at 21 16 02

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,6 +44,7 @@
         "just-debounce-it": "^3.2.0",
         "memoize-one": "^6.0.0",
         "postcss": "^8.4.40",
+        "postcss-assign-layer": "^0.4.0",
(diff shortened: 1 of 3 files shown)


skoch13 (approved):
🏋️🚀

andrey-skl:
Note: we faced an issue with layers approach that it is not possible to override styles partially, see https://codepen.io/huston007/pen/BagwMEe?editors=1100

We've decided to freeze this pull request until we maybe understand how to deal with that

PR #7835, opened 2024-10-21
chore(eslint): migrate to eslint@9

skoch13:
Migrated current configuration to `eslint9`. Points to mention:

* I needed to add `prettier` and its recommended rules because, starting from `eslint8`, styling was decoupled from the linter. We were using deprecated rules, which I removed, and I added the opinionated `prettier/recommended` rules.
* I had to mute `no-lone-blocks` for `src/select/select.stories.tsx` as we clearly have issues there, but I'm not sure I get the initial idea of blocks. So I ask @Hypnosphi to help me with that
* I needed to add `@babel/plugin-syntax-import-assertions` to support the `with/assert` syntax before `eslint` provides its own support. (`with` syntax just went to stage 4)
* @Hypnosphi please also help me with `packages/screenshots/testplane/test.mjs` as it has reference to a non-existent file `error  Unable to resolve path to module './stories.json'  import/no-unresolved`

https://youtrack.jetbrains.com/issue/RG-2503


Hypnosphi:
> I had to mute no-lone-blocks for src/select/select.stories.tsx as we clearly have issues there, but I'm not sure I get the initial idea of blocks. So I ask @Hypnosphi to help me with that

consts like `avatarUrl` and `tags` will clash between different stories if you remove the blocks. Also, the rule description says that usage with `const` shouldn't be an error, so maybe we have some misconfiguration here
https://eslint.org/docs/latest/rules/no-lone-blocks
> In ES6, code blocks may create a new scope if a block-level binding (let and const), a class declaration or a function declaration (in strict mode) are present. A block is not considered redundant in these cases.

> @Hypnosphi please also help me with packages/screenshots/testplane/test.mjs as it has reference to a non-existent file error  Unable to resolve path to module './stories.json'  import/no-unresolved
This file is generated in runtime so it's OK to suppress the rule for this particular line

Hypnosphi (approved):

skoch13:
> so maybe we have some misconfiguration here

you're right, `ts(x)` files were transpiled to es2015, adjusting this resolved the issue.

vkrol (approved):

andrey-skl:
@skoch13 thanks a ton!

PR #7841, opened 2024-10-22
chore(git): add .git-blame-ignore-revs

skoch13:
Small utility change: within the codebase there might be commits that one wants to ignore for `blame` such as recent `apply prettier suggestions`. See before 
image
and after
image


andrey-skl (approved):
Thanks, I also thought about asking you to bring this sweat-neat improvement, but you already did it 👍

PR #7853, opened 2024-10-24
Drop index type from SelectionItem to make the typings more strict

andrey-skl:
For example, this is what we have to do to extend `SelectionItem` without this fix:

```ts
export interface CustomAppInTableSelection extends CustomAppLite {
  // hack for TableSelection type mismatching
  [k: string]: unknown;
}

...

selection: TableSelection<CustomAppInTableSelection>;
```


skoch13 (approved):
Great! Thank you!

PR #7857, opened 2024-10-25
RG-2504 fix CSS order issue in ring-ui-built

andrey-skl:


Hypnosphi:
I'm not familiar with the built part so I can't really give any input here

vkrol (approved):

PR #7868, opened 2024-10-28
Reduce usage of postcss-modules-values-replace `@value` syntax for future plugin removal

andrey-skl:
I think it makes sense to reduce usage of non-standard CSS syntax to bare minimum

Therefore I replaced local usages of `@value` syntax with CSS Custom Properties

`@value` is still used to import hashed classNames from other CSS files, but this is covered by `postcss-modules-values` [plugin](https://github.com/madyankin/postcss-modules/blob/bd64c71ddfd81b615104b0727ce9a623da0eaef6/src/scoping.js#L6) and is a part of CSS Modules stack

### DIFF
--- a/src/grid/grid.css
+++ b/src/grid/grid.css
@@ -2,35 +2,36 @@
 
 @value breakpoint-small, breakpoint-middle, breakpoint-large  from "../global/global.css";
 @value large-screen-media, middle-screen-media, small-screen-media from "../global/global.css";
-@value gutterWidth: (var(--ring-unit) * 2);
-@value gutterCompensation: calc(gutterWidth / -2);
-@value outerMargin: calc(var(--ring-unit) * 2);
-@value containerSmall: calc(breakpoint-small + gutterWidth);
-@value containerMedium: calc(breakpoint-middle + gutterWidth);
-@value containerLarge: calc(breakpoint-large + gutterWidth);
-@value width-1: 8.3333%;
-@value width-2: 16.6667%;
-@value width-3: 25%;
-@value width-4: 33.3333%;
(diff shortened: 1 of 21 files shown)


vkrol (approved):

src/grid/grid.css
```diff
@@ -2,35 +2,36 @@
...
+  --ring-grid-width-4: 33.3333%;
+  --ring-grid-width-5: 41.6667%;
+  --ring-grid-width-6: 50%;
+  --ring-grid-width-7: 58.3333%;
+  --ring-grid-width-8: 66.6667%;
+  --ring-grid-width-9: 75%;
+  --ring-grid-width-10: 83.3333%;
+  --ring-grid-width-11: 91.6667%;
+  --ring-grid-width-12: 100%;
+
```
Hypnosphi, line 26: @andrey-skl restricting those to the .container scope is a breaking change if you use Row outside of Grid. It should be at least mentioned in the changelog but a safer option would be to define them all on `:root` level, for all the components

andrey-skl: Thanks! Didn't think that `<Row>` may be useful outside of grid

Hypnosphi: I'm not sure we actually use it that way, maybe it's just storybook examples. So maybe `:root` would be an overkill because it would bloat the devtools

andrey-skl: I would also vote for keeping them on `.container` level to avoid polluting global variables namespace

andrey-skl: > Maybe something like this could work better:

I like that, let's do so, thanks for the idea

PR #7906, opened 2024-11-08
Fix Grid variables not available for orphan <Row/>, see #7868

andrey-skl:


Hypnosphi (changes requested):

src/grid/grid.css
```diff
@@ -4,7 +4,8 @@
 @value large-screen-media, middle-screen-media, small-screen-media from "../global/global.css";
 
 .container-fluid,
-.container {
+.container
+.row {
```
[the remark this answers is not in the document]
Hypnosphi: Although I double-checked and we never use `Col` without having `Row` in the same component, so it should be fine as is

Hypnosphi (approved):

src/grid/grid.css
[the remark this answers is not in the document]
andrey-skl: Thanks for checking!

src/grid/grid.css
```diff
@@ -4,7 +4,8 @@
 @value large-screen-media, middle-screen-media, small-screen-media from "../global/global.css";
 
 .container-fluid,
-.container {
+.container
```
Hypnosphi, line 7: a comma is missing here, so it's `.container .row` instead of `.container, .row`

Hypnosphi: I'll fix that

andrey-skl: Ouch I'm sorry, thank you for your help!

PR #7944, opened 2024-11-19
Separate hover and selected states for list items Fix RG-2179 Press [Enter] select an item hovered by mouse cursor accidentally

Hypnosphi:


andrey-skl (approved):
Looks great, thanks!

PR #7945, opened 2024-11-19
RG-2519 Tabs: allow plain links without onSelect

Hypnosphi:


andrey-skl (approved):
Thanks a lot for taking this!

PR #8039, opened 2024-12-16
RG-2525 feat(DropdownMenu): accept children

skoch13:
So it makes it possible to access `popupProps` and call methods from it, e.g. `onCloseAttempt`.

Case: when one renders `DropdownMenu` inside of a `DropdownMenu` and wants to close the parent `DropdownMenu` on the children's select. Otherwise consumer have to stop event propagation on the wrapper level, and create hackish solutions such as `DropdownMenu` key re-generation on `onSelect` calls.  Example in story.

### DIFF
--- a/src/dropdown-menu/dropdown-menu.tsx
+++ b/src/dropdown-menu/dropdown-menu.tsx
@@ -1,7 +1,7 @@
 import {forwardRef, useMemo, cloneElement, ReactElement, HTMLAttributes, SyntheticEvent, Ref, ReactNode} from 'react';
 
 import List, {ActiveItemContext, SelectHandlerParams} from '../list/list';
-import Dropdown, {AnchorProps, DropdownAttrs, DropdownChildren} from '../dropdown/dropdown';
+import Dropdown, {AnchorProps, DropdownAttrs, DropdownChildrenFunction} from '../dropdown/dropdown';
 import PopupMenu, {PopupMenuAttrs} from '../popup-menu/popup-menu';
 import {PopupAttrs} from '../popup/popup';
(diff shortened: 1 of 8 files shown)


andrey-skl (approved):

src/dropdown-menu/dropdown-menu.tsx
```diff
@@ -67,23 +72,52 @@ function DropdownAnchorWrapper({
   return <div {...anchorAriaProps}>{anchor}</div>;
 }
 
+function renderDropdownMenuChildren<T>({children, popupMenuProps}: DropdownMenuChildren<T>) {
+  const {data, ...restPopupMenuProps} = popupMenuProps;
```
Hypnosphi, line 76: Why exclude data?

skoch13: otherwise we overwrite `data`  provided to `PopupMenu`. Do you suggest excluding `data` from props on a consumer level?

Hypnosphi: If user provides `data` directly to `DropdownMenu` we shouldn't just ignore it.

> we overwrite data provided to PopupMenu

We don't overwrite anything if `children` is a function. If it's a React element, you can prevent `cloneElement` resetting `data` to undefined by checking that it's not undefined

Hypnosphi: Alternatively, you can make it so that the inner props always take precedence:

```
cloneElement(children, {...popupMenuProps, ...children.props})
```

Hypnosphi (approved):

PR #8062, opened 2024-12-20
Merge in release 6.x into master

andrey-skl:


skoch13 (approved):

packages/screenshots/testplane/test.mjs
```diff
@@ -5,7 +5,6 @@ import filenamify from 'filenamify';
 
 import Actions from './actions.js';
 
-// eslint-disable-next-line import/no-unresolved
```
Hypnosphi, line 8: this is a generated file that's absent until you run visual tests

andrey-skl: So I should delete it? Sure, thanks!

Hypnosphi: Yes, and it means that we still need a eslint-disable here

andrey-skl: Fun part is that one have to hack this around : D

Hypnosphi (approved):

PR #8092, opened 2025-01-06
RG-2550 fix tooltip hides if contains complex markup

andrey-skl:


skoch13:
Very cool, thank you. I'm just wondering if we still need the handler from lines 136-142 ?

andrey-skl:
> Very cool, thank you. I'm just wondering if we still need the handler from lines 136-142 ?

Unfortunatelly yes I'm afraid. They handle `mouseout` on the anchor, while the newly added "hideIfMovedOutsidePopup" handles `mouseout` for popup. I am not sure I can see a way to unite them

skoch13:
@andrey-skl I tried removing it and nothing broke. could you please double-check?

skoch13 (approved):

andrey-skl:
> @andrey-skl I tried removing it and nothing breaks. could you please double-check?

Sure, thanks for trying, sounds interesting!

andrey-skl:
@skoch13 If I comment the condition the original issue [RG-2490](https://youtrack.jetbrains.com/issue/RG-2490) is back:

skoch13:
> @skoch13 If I comment the condition the original issue [RG-2490](https://youtrack.jetbrains.com/issue/RG-2490) is back

`this.hidePopup()` call might be also omitted here as `onMouseOut` will call it when expected

andrey-skl:
> `this.hidePopup()` call might be also omitted here as `onMouseOut` cover all the required cases when expected. So, I removed the `mouseout` event handler completely during my tests.

Got it! But then leaving the anchor to the side won't close the popup, see:

PR #8128, opened 2025-01-15
RG-2556 fix(List): remove hover styles when custom item is disabled

ks-ieremenko:
`.error` className seems to be unused, so I removed it. Please check that I haven't accidentally missed it

### DIFF
--- a/src/list/list.css
+++ b/src/list/list.css
@@ -78,18 +78,6 @@
   line-height: calc(var(--ring-unit) * 2);
 }
 
-.error {
-  cursor: default;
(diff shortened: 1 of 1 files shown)


andrey-skl:
@ks-ieremenko thanks for the PR! It would need for me to checkout the branch and test a bit, so I'd take a look tomorrow

Hypnosphi:
Unfortunately, this broke hover for link items but https://github.com/JetBrains/ring-ui/pull/8139 will fix it

PR #8222, opened 2025-02-14
RG-2572 Update tooltips to match design styles

Hypnosphi:


src/tooltip/tooltip.tsx
```diff
@@ -26,7 +26,7 @@ export interface TooltipProps extends Omit<AllHTMLAttributes<HTMLSpanElement>, '
   selfOverflowOnly?: boolean | null | undefined;
   popupProps?: Partial<PopupAttrs> | null | undefined;
   title?: ReactNode | null | undefined;
-  theme?: Theme;
+  theme?: Theme | 'inherit';
```
Hypnosphi, line 29: I'm not sure whether it's better to do it like that or have a separate prop `inheritTheme`

andrey-skl: Looks good!

PR #8339, opened 2025-03-19
Take scrollbar-gutter into account in dialog body scroll preventer

vkrol:


andrey-skl:
I actually think this scroll preventer should be replaced with overscroll-behavior: [contain](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior#contain). But I do understand that this might be not the best moment to implement

andrey-skl (approved):

vkrol:
`overscroll-behavior: contain` will not allow achieving the same effect, so in my opinion, it is not enough.

PR #8371, opened 2025-03-28
Fix List shouldComponentUpdate

vkrol:
`nextProps !== this.props` is always `true` in React 19, so `setState` in `componentDidUpdate` results in an infinite loop. I also use `Object.is` because the `maxHeight` prop can contain `NaN`, `NaN` !== `NaN`.


andrey-skl:
Do we often compare props like that? Maybe we need some generic helper

andrey-skl (approved):

vkrol:
We have only 3 places where shouldComponentUpdate is used, and in all three the logic is different.

PR #8387, opened 2025-04-03
Use react 19

andrey-skl:


vkrol (approved):

Hypnosphi (approved):

andrey-skl:
I'm having issue with Input's `autoFocus` [attribute](https://github.com/JetBrains/ring-ui/blob/4e172a0aa40cfe1e5725b83ffed099f5796cbe57/src/select/select__filter.tsx#L65) not working for Select dropdown if `<StrictMode>` is enabled:


cc @vkrol I recall you faced similar issue in YouTrack Inline Comments recently

vkrol:
@andrey-skl maybe we can revert that commit https://github.com/JetBrains/ring-ui/commit/8e0f12d0a43617850bd4bdf45295d739ba145ef6. What do you think?

PR #8517, opened 2025-05-23
Built version: use human readable classnames [publish]

Hypnosphi:


andrey-skl:
@katriyna Could you please also review this, as you use built version I believe?

andrey-skl:
We've agreed to wait a bit while @katriyna can check if this version works fine

Published as [7.0.48-beta.0](https://teamcity.jetbrains.com/buildConfiguration/JetBrainsUi_RingUi_PublishNext/5205889)

andrey-skl:
@katriyna says it works fine, so let's merge? @Hypnosphi

PR #8524, opened 2025-05-26
fix: add explicit extension for dynamic languages import

skoch13:


skoch13:
For vite it still doesn't work without [the plugin](https://www.npmjs.com/package/vite-plugin-dynamic-import)
Screenshot 2025-05-26 at 14 42 58

With it and proper configuration Code component start to work:

```js
    dynamicImport({
      filter: () => true,
      onResolve: (importee: string) => {
        if (importee.includes("language")) {
          return importee;
        }
        return undefined;
      },
    }),
```

PR #8591, opened 2025-06-12
RG-2197 fix auto highlight for multi select

lampmaster:
The problem occurred because activeIndex (highlighted item) was calculated before the reset button and separator were added to the list.

As a result, the highlighted item appears two positions earlier than the actual selected one (see video 1)

https://github.com/user-attachments/assets/0ea98a40-582a-4576-b90a-f0a1b31dd84f

**Current behavior (video 2):**
- Keys are now used instead of indexes to determine which item to highlight — this is safer and more stable 
- Only actual option items (not separators or reset buttons) can be auto-highlighted when the popup opens
- The last selected/unselected element is highlighted after interaction
- In multiselect mode, the last selected tag is highlighted when the popup opens

This behavior now matches the UX of other UI kits such as [Material UI](https://mui.com/material-ui/react-select/#multiple-select), providing a more consistent and predictable experience

https://github.com/user-attachments/assets/804313cb-791c-4978-8bb0-7e675b323647


andrey-skl:
Looks great, thanks for a thoroughful thinking through the issue!

PR #8635, opened 2025-06-23
RG-2599 exclude Dialog from autodocs

lampmaster:
Now in Dialog and ErrorBubbles storyes with Dialog hide other stories

It's more convenient to interact with the Dialog component through separate stories
So it looks like the easiest solution is to exclude Dialog from autodocs

An example in the Dialog component 

https://github.com/user-attachments/assets/f3e47dcf-495b-498b-b3fa-72140239bf44


andrey-skl:
Great idea, thanks!

PR #8661, opened 2025-06-27
RG-2586 Indeterminate Checkbox for Data List

amanokh:


src/data-list/data-list.tsx
```diff
@@ -130,9 +132,10 @@ class DataList<T extends SelectionItem> extends PureComponent<DataListProps<T>>
                 onExpand={item.onExpand}
                 showFocus={selection.isFocused(model)}
                 onFocus={this.onItemFocus}
-                selection={selection}
+                selection={selection as Selection<T>}
```
amanokh, line 135: Actually, it is not so good to make type-assertions. However, otherwise we need to fully change the interface of `selection-shortcuts-hoc.tsx` by adding a new generic type for Selection (DataListSelection / TableSelection), which may be a breaking change for a hook's end-users :(

vkrol (approved):

PR #8693, opened 2025-07-09
RG-2472 When hovering over Remove button of Tag, Tag itself loses hovered state

amanokh:


src/tag/tag.tsx
```diff
@@ -112,18 +110,27 @@ export default class Tag extends PureComponent<TagProps> {
...
+    this.props.onRemove?.(e);
+    e.stopPropagation();
+    e.preventDefault();
+  };
+
   renderRemoveIcon() {
     if (!this.props.readOnly && this.props.onRemove) {
       return (
-        <Button
+        <Icon
```
andrey-skl: I'm afraid this line can be problematic, making Icon interactive removes all the accessibility capabilities that Button have. Also it could not be focused by Tab. Is it strictly necessary to use `Icon`?

amanokh: The problem is that it is prohibited to have `<button>` as a descendant of `<button>`. 
image

For accessibility work, I've additionally handled tabIndex and onKeyDown

andrey-skl: Thanks! I understand the reason now

Before continue with this approach, haven't you tried to move the :hover styles up to the .container?

Screenshot 2025-07-09 at 20 43 15

That way, probably, hover would work for any element inside container

amanokh: thanks, that worked way much better)

andrey-skl (approved):

PR #8702, opened 2025-07-14
colors

Hypnosphi:


CONTRIBUTING.md
```diff
@@ -70,3 +70,13 @@ To check current status on CI you may check [teamcity configuration](https://tea
 
 To run tests locally, run `npm run a11y-audit`. Also, there is "Accessibility" tab on storybook pages, 
 so every component could be inspected via running storybook (`npm start`) and then checking this tab.
+
+### Updating color variables from Figma
+
+- Go to https://www.figma.com/design/HY6d4uE1xxaQXCMG9fe6Y2/RingUI--Internal-?m=auto&node-id=1-58
+- In "Plugins" tab, find and run "variables2css"
+- Select "Theme/Light" and "Theme/Dark"
```
amanokh, line 78: Should we also store primitives in repo and add a step to update them maybe?

design-tokens/update-styles.mjs
```diff
@@ -0,0 +1,50 @@
...
+
+for (const {mode, color} of tokens[0].values) {
+  const filename = `src/global/variables${mode.name === 'Dark' ? '_dark' : ''}.css`;
+  const source = fs.readFileSync(path.resolve(__dirname, `../${filename}`), 'utf-8');
+  let result = source;
+  const newStyles = [];
+  for (const {name, value} of color) {
+    if (/^ring-.*-color$/.test(name)) {
+      const components = name.replace('-color', '-components');
+      const {r, g, b} = hexToRgb(value);
```
amanokh, line 32: We can export colors as rgb right from the plugin by applying this config 
image

design-tokens/update-styles.mjs
```diff
@@ -0,0 +1,50 @@
...
+      }
+    }
+  }
+  if (result !== source) {
+    fs.writeFileSync(path.resolve(__dirname, `../${filename}`), result);
+  }
+  if (newStyles.length > 0) {
+    // eslint-disable-next-line no-console
+    console.log(`New variables detected. Please add them to ${filename}:\n${newStyles.join('\n')}`);
+  }
```
amanokh, line 49: Can we add the new variables to the file right away by this script?

design-tokens/update-styles.mjs
```diff
@@ -0,0 +1,50 @@
...
+    if (/^ring-.*-color$/.test(name)) {
+      const components = name.replace('-color', '-components');
+      const {r, g, b} = hexToRgb(value);
+      if (!source.includes(`--${name}:`)) {
+        newStyles.push(`  --${components}: ${r}, ${g}, ${b};`);
+        newStyles.push(`  --${name}: rgb(var(--${components})); /* ${value.toUpperCase()} */`);
+      } else {
+        result = result
+          .replace(new RegExp(`--${components}: .+;`), `--${components}: ${r}, ${g}, ${b};`)
+          .replace(new RegExp(`(--${name}: .+;) /\\* .+? \\*/`), `$1 /* ${value.toUpperCase()} */`);
```
amanokh, line 39: Why we store rgb colors separately as 3-numbers array in `--x-components` variable, but not directly as rgb(...) color in `--x-color`?

design-tokens/update-styles.mjs
```diff
@@ -0,0 +1,50 @@
...
+
+import {fileURLToPath} from 'node:url';
+
+import tokens from './design-tokens.json' with {type: 'json'};
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+// https://stackoverflow.com/a/5624139
+function hexToRgb(hex) {
```
amanokh, line 13: This won't work for colors with alpha channel

design-tokens/update-styles.mjs
Hypnosphi: we need both rgb (for `-components`) and hex (for comment)

design-tokens/update-styles.mjs
Hypnosphi: @katriyna introduced it at some point, to be able to easily add alpha to any color

design-tokens/update-styles.mjs
Hypnosphi: Luckily our designers don't use those. If they start to, we can add some additional checks

design-tokens/update-styles.mjs
Hypnosphi: no, because it's not clear in what block to add them

CONTRIBUTING.md
Hypnosphi: I don't think it's a good idea, because they shouldn't be exposed to the developers

amanokh (approved):

PR #8705, opened 2025-07-16
fix: do not render pager if total is less than page size

ks-ieremenko:
If total is less than page size and page selector is disabled, then pager is rendered as an empty div with class names.

### DIFF
--- a/src/pager/pager.tsx
+++ b/src/pager/pager.tsx
@@ -315,10 +315,15 @@ export default class Pager extends PureComponent<PagerProps> {
 
   render() {
     const classes = classNames(style.pager, this.props.className);
+    const shouldRenderPagerContent = this.getTotalPages() > 1 || this.props.openTotal;
+
+    if (!shouldRenderPagerContent && this.props.disablePageSizeSelector) {
+      return null;
+    }
 
     return (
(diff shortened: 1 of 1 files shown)


src/pager/pager.tsx
```diff
@@ -315,10 +315,15 @@ export default class Pager extends PureComponent<PagerProps> {
...
+    const isSinglePageOrEmpty = this.getTotalPages() <= 1;
+
+    if (isSinglePageOrEmpty && !this.props.openTotal && this.props.disablePageSizeSelector) {
+      return null;
+    }
 
     return (
       <div data-test="ring-pager" className={classes}>
-        {this.getTotalPages() > 1 || this.props.openTotal ? this.getPagerContent() : this.getPageSizeSelector()}
+        {!isSinglePageOrEmpty || this.props.openTotal ? this.getPagerContent() : this.getPageSizeSelector()}
```
Hypnosphi: To avoid possible desynchronization of the conditions, let's maybe extract this whole line into a variable and just check that it's not empty in the condition on line 320

ks-ieremenko: I changed the logic a bit; should work the same way. Please take a look 🙏

PR #8722, opened 2025-07-22
RG-2633 Remove unnecessary description margin in Confirm component

ks-ieremenko:
There are already paddings on `Header` and `Content` components, so in this case description margin creates unnecessary offset.


ks-ieremenko:
~~The only thing: I don't have BrowserStack credentials, so can you please help with screenshots?🙏~~ updated

andrey-skl (approved):

PR #8827, opened 2025-09-01
refactor(teamcity): refactor TeamCity configs, update Qodana version

amanokh:
For usability to edit configs in IntelliJ IDEA, I've split the build configs into separated files. This enables syntax highlighting, completion e.t.c.
image

Also I've updated Qodana version, seems it fixed some issues with configuration, the number of problems reported decreased
image


amanokh:
@Hypnosphi it seems we also should disable editing the TC configs through UI, because it actually doesn't change the DSL configs in VCS, but creates a copy of them into patches that may brake things :(

Hypnosphi:
> it seems we also should disable editing the TC configs through UI, because it actually doesn't change the DSL configs in VCS, but creates a copy of them into patches that may brake things :(

I would really like not to remove that ability

PR #8839, opened 2025-09-03
refactor: refactor some root folders/scripts

amanokh:


andrey-skl (approved):

maxmaximov:
Folks, these changes break "Publish to GitHub" build configuration as it tries to use `chromatic` which was deleted: https://github.com/JetBrains/ring-ui/blob/4a430ddca07453f16c2bb98d4b62167b81790bb0/.teamcity/src/publish/PublishToGitHubPages.kt#L75

Apperantly we need to update that script?

amanokh:
@maxmaximov Thanks, fixed [#8847](https://github.com/JetBrains/ring-ui/pull/8847)

PR #8860, opened 2025-09-08
RG-2321 implement global "configure()" method

andrey-skl:


src/global/controls-height.tsx
```diff
@@ -1,20 +1,18 @@
...
-}
+import deprecate from 'util-deprecate';
+
+import {configure, ControlsHeight, getConfiguration} from './configuration';
 
-export const ControlsHeightContext = createContext(ControlsHeight.M);
+export {ControlsHeight} from './configuration';
 
-let globalControlsHeight = ControlsHeight.M;
+export const ControlsHeightContext = createContext<ControlsHeight | (() => ControlsHeight)>(ControlsHeight.M);
```
Hypnosphi, line 9: I think it should use `getGlobalControlsHeight` as the default context value

Hypnosphi: I think it should use `getGlobalControlsHeight` as the default context value

Hypnosphi (changes requested):

src/global/controls-height.tsx
andrey-skl: Right, missed that. Thanks!

PR #8863, opened 2025-09-10
RG-2648 - Progress bar refactor 

lampmaster:


andrey-skl:
Given that this is a reference case and we'll be using it as a reference in next refactorings, let's review it alltogether ;)

src/progress-bar/progress-bar.interface.ts
```diff
@@ -0,0 +1,33 @@
...
+   */
+  max?: number;
+  /**
+   * A floating point number that specifies current progress value.
+   * @type {number}
+   */
+  value?: number;
+  label?: string;
+  /**
+   * Sets the ring-progress-bar_global class to position the progress bar on top of the screen.
```
Hypnosphi: I think it's `ring-progress-bar-globalMode`

src/progress-bar/progress-bar.stories.tsx
```diff
@@ -1,51 +1,102 @@
...
+};
+
+export default meta;
+type Story = StoryObj<typeof ProgressBar>;
+
+export const Basic: Story = {};
+
+export const Animated: Story = {
+  render: args => {
+    const AnimatedDemo = () => {
```
Hypnosphi: While we're at it, let's extract all component definitions to the module scope

src/progress-bar/progress-bar.tsx
```diff
@@ -85,7 +38,9 @@ export default class ProgressBar extends PureComponent<ProgressBarProps> {
...
       </div>
     );
-  }
-}
+  },
+);
+
+ProgressBar.displayName = 'ProgressBar';
 
-export type ProgressBarAttrs = React.JSX.LibraryManagedAttributes<typeof ProgressBar, ProgressBarProps>;
```
Hypnosphi, line 91: Please document the removal of this export in `CHANGELOG.md`

src/progress-bar/progress-bar.tsx
```diff
@@ -1,81 +1,34 @@
...
-  progressbarRef = (el: HTMLElement | null) => {
-    this.progressbar = el;
-  };
-
-  render() {
-    const {className, global, max, value, label, staticColor, ...otherProps} = this.props;
-
-    const width = value ? `${ProgressBar.toPercent(value, max)}%` : undefined;
+import {toPercent} from './progress-bar.utils';
+import {ProgressBarProps} from './progress-bar.interface';
```
Hypnosphi: let's reexport it here

src/progress-bar/progress-bar.utils.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Converts progress value to percentage
+ * @param {number} value The progress task value
+ * @param {number} max The maximum value
+ * @return {number} The progress task value in percent
+ */
+export const toPercent = (value: number, max: number): number => {
+  const HUNDRED_PERCENT = 100;
+  const percents = Math.max((value * HUNDRED_PERCENT) / max, 0);
+  return percents > HUNDRED_PERCENT ? HUNDRED_PERCENT : percents;
```
amanokh: Let's make common `clamp` function and store it in src/utils?

src/progress-bar/progress-bar.tsx
```diff
@@ -1,81 +1,34 @@
...
+      label = 'Progress',
+      staticColor = false,
+      disableAnimation = false,
+      global,
+      className,
+      ...otherProps
+    },
+    ref,
+  ) => {
+    const width = value ? `${toPercent(value, max)}%` : undefined;
```
amanokh:
```suggestion
    const width = value && `${toPercent(value, max)}%`;
```

src/progress-bar/progress-bar.interface.ts
lampmaster: Yeah, you're right. And I think it's even better to use clearer description without technical implementation details.

src/progress-bar/progress-bar.tsx
lampmaster: Yeah, your are right. How about reexport components entities from index.ts and hide internal utils (like toPercent) from external users?

Hypnosphi: We used to have an index.ts but we weren't able to make tree shaking work properly. But you can try again

lampmaster: Ohhh, thanks. It looks like it's better to create a separate issue for this and for now implement your suggestion. What do you think?

lampmaster: https://youtrack.jetbrains.com/issue/RG-2659/Unify-component-exports-with-index.ts

Hypnosphi: Sure. My point here is not to break the existing imports when possible. To be clear, I'm just talking about `ProgressBarProps` because it was exported from this file before

lampmaster: Thanks)

Hypnosphi: Unresolving as it's not fixed yet

src/progress-bar/progress-bar.interface.ts
```diff
@@ -0,0 +1,33 @@
+import {HTMLAttributes} from 'react';
+
+export interface ProgressBarProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * A floating point number that specifies maximum progress value. Default value is 1.0.
+   * @type {number}
+   */
+  max?: number;
```
amanokh: please put readonly to all props

src/progress-bar/progress-bar.stories.tsx
```diff
@@ -1,51 +1,102 @@
...
+  render: args => (
+    <div>
+      <p>This progress bar is positioned globally at the top of the viewport:</p>
+      <ProgressBar {...args} global label="Global progress" />
+    </div>
+  ),
+  parameters: {
+    screenshots: {skip: true},
+  },
+};
```
amanokh, line 107: Could you please make test-only stories for value=0 state also?

src/progress-bar/progress-bar.tsx
lampmaster: Fixed and checked locally import type {ProgressBarProps} from '@jetbrains/ring-ui/components/progress-bar/progress-bar'

src/progress-bar/progress-bar.utils.ts
lampmaster: Yeah, sure thanks)

src/progress-bar/progress-bar.interface.ts
```diff
@@ -0,0 +1,31 @@
...
+   * A floating point number that specifies current progress value.
+   * @type {number}
+   */
+  readonly value?: number;
+  readonly label?: string;
+  /**
+   * Positions the progress bar globally at the top of the screen.
+   * @type {boolean}
+   */
+  readonly global?: boolean | null | undefined;
```
amanokh, line 19: I think we should discontinue usage of `null`, let's remove it from possible types and leave only `global?: boolean`?
(It would also be a breaking change a bit, so better to put this to changelog)

Hypnosphi: I think we shouldn't overly restrict the interfaces. I can have something like this and it should work perfectly fine:

```tsx
const myCustomFlag: boolean | null

<ProgressBar global={myCustomFlag} />
```

if we forbid `null` I'd have to add an extra conversion to boolean for no good reason

lampmaster: So let's leave it as it's? @amanokh @Hypnosphi

lampmaster: Actually I thought about it a bit more and I agree with Anton. A strict interface looks clearer. With boolean | null it’s not obvious what the difference is between null and undefined, which may confuse users of the API.

Hypnosphi: Why there should be any difference?

PR #8879, opened 2025-09-15
refactor(eslint): RG-2636 Refactor ESLint config, enforce more strict rules

amanokh:
What was mainly changed:
- eslint config was refactored to use common `languageOptions`
- enforced kebab-style (like `name-of-component.utils.tsx`) to filenames in /src
- enforced singlequote strings
- enforced consistent type definitions (type/interface)
- fixed prettier config
- enforced usage of `=== `instead `==`, all occurrences of checking the `value != null` were changed either to `!!value` or `value !== null && value !== undefined` (the second is for cases when the value could be an empty string or 0, this should be rewritten in future to use only `undefined` for types)
- added max-lines=400 limit for all files
- enforced typed imports
- added some stylelint rules
- added autofixing on save for WebStorm configs


eslint.config.mjs
```diff
@@ -1,77 +1,85 @@
...
+  // Global ignores
   {
     ignores: [
+      'components',
+      'docs',
       '**/coverage',
       '**/dist',
-      '**/components',
       '**/test-app',
-      '**/storybook-dist',
```
Hypnosphi, line 45: this dir is created when you run `npm run build-stories`, it shouldn't be linted

eslint.config.mjs
```diff
@@ -80,216 +88,341 @@ export default tseslint.config(
...
+      },
     },
 
     rules: {
-      'valid-jsdoc': 0,
-      'import/no-commonjs': 0,
-
-      'consistent-return': 0, // as we rely on tsconfig's noImplicitReturns
-
+      'array-callback-return': 'error',
```
Hypnosphi: this is already present in `@jetbrains/eslint-config` https://github.com/JetBrains/eslint-config/blob/8638cbb520e4d26ea380137b3897909c2f7edfd2/base.js#L33

please check other rules too

Hypnosphi (approved):

src/analytics/analytics-custom-plugin.ts
```diff
@@ -82,7 +82,7 @@ export default class AnalyticsCustomPlugin implements AnalyticsPlugin {
   private _addDataToFlushingPack(sendingData: AnalyticsCustomPluginData) {
     this._data.push(sendingData);
 
-    if (this._flushMaxPackSize != null && this._data.length >= this._flushMaxPackSize) {
+    if (this._flushMaxPackSize && this._data.length >= this._flushMaxPackSize) {
```
Hypnosphi, line 85: I think flushMaxPackSize can be 0 as well

@amanokh can you please bring back the checks for null and undefined in all places where you removed it?

PR #8881, opened 2025-09-19
use secondary color for empty select state, unless it's disabled

Hypnosphi:


.storybook/preview.ts
```diff
@@ -21,10 +20,7 @@ if (params.has('block-animations')) {
   document.body.classList.add(styles.blockAnimations);
 }
 
-const selectorsWithColorContrastIssues = [
-  `.${queryAssistStyles.placeholder.split(' ')[0]}`,
-  `.${selectStyles.buttonValueEmpty}`,
-].filter(Boolean);
+const selectorsWithColorContrastIssues = [`.${queryAssistStyles.placeholder.split(' ')[0]}`].filter(Boolean);
```
amanokh: Do we still need an array filter here?

amanokh (approved):

.storybook/preview.ts
Hypnosphi: not really

PR #8883, opened 2025-09-19
RG-2660 Update disabled active color

Hypnosphi:


packages/screenshots/testplane/chrome/components/checkbox/basic/basic-dark.png
amanokh, line 1: I'm not sure whether the disabled checkbox mark should have such low-contrast grey color — in Figma mockups it is different 
Screenshot 2025-09-22 at 10 56 59

amanokh:
In mockups, the --ring-border-hover-color is actually used for disabled background, not --ring-border-disabled-active-color (besides, there is no such token in the Figma library)
Screenshot 2025-09-22 at 11 13 27

src/global/variables.css
amanokh, line 1: In mockups, the --ring-border-hover-color is actually used for disabled background, not --ring-border-disabled-active-color (besides, there is no such token in the Figma library)
Screenshot 2025-09-22 at 11 13 27

amanokh (changes requested):

packages/screenshots/testplane/chrome/components/checkbox/basic/basic-dark.png
amanokh: p.s. I'm not sure whether the white mark color in Figma mockup is correct too, though 😅 It doesn't look like «disabled»
Could you re-check it with designers, please?

Hypnosphi: Must be an error on Figma, this state isn't represented on any of the screens so the designers probably missed it. I'll report it to Ryta

Hypnosphi: I'll change it to `--ring-secondary-color`, same as with primary button

andrey-skl (approved):

PR #9261, opened 2026-04-10
RG-2773 Fix empty row on scroll up in dropdown

Hypnosphi:
## Summary
- Use `IntersectionObserver` on the List container to detect invisible→visible transitions and call `recomputeRowHeights()`, fixing stale row height cache from `react-virtualized` that caused a phantom empty row when scrolling up in a dropdown
- Mark the unused `visible` prop as `@deprecated` for removal in Ring UI 8.0

## Test plan
- [x] Run existing list tests (`npx jest src/list/`)
- [x] Run select tests (`npx jest src/select/`)
- [x] Manual: open a dropdown with a long list, scroll down, close, reopen, scroll up — no empty row should appear

### DIFF
--- a/src/list/list.tsx
+++ b/src/list/list.tsx
@@ -96,6 +96,9 @@ export interface ListProps<T = unknown> {
   maxHeight?: number | null | undefined;
   activeIndex?: number | null | undefined;
   useMouseUp?: boolean | null | undefined;
+  /**
+   * @deprecated No longer used. Visibility is detected automatically via IntersectionObserver. Will be removed in Ring UI 8.0.
+   */
   visible?: boolean | null | undefined;
   disableMoveOverflow?: boolean | null | undefined;
   compact?: boolean | null | undefined;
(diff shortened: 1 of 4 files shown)


src/list/list.tsx
```diff
@@ -212,6 +215,17 @@ export default class List<T = unknown> extends Component<ListProps<T>, ListState
     if (!this.props.renderOptimization) {
       this.scrollToActive();
     }
+    if (this.props.renderOptimization && this.container && typeof IntersectionObserver !== 'undefined') {
+      this.intersectionObserver = new IntersectionObserver(entries => {
+        for (const entry of entries) {
+          if (entry.isIntersecting && !this.wasVisible && this.virtualizedList) {
+            this.virtualizedList.recomputeRowHeights();
+          }
+          this.wasVisible = entry.isIntersecting;
```
aleksei-berezkin, line 224: Patching this field for every `entry` seems a bit misleading (until you look below on the `observe()` call). Maybe deliberately only take the first entry? I.e. `new IntersectionObserver(([container]) => ...)`

aleksei-berezkin (approved):

src/list/list.tsx
Hypnosphi: https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver#entries

> You should not assume the number of entries, because multiple threshold-crossing events may be reported in a single callback invocation

PR #9306, opened 2026-06-08
RG-2542 table -> legacy-table

aleksei-berezkin:
This PR moves all the existing tables:

- In file system `src/table` -> `src/legacy-table`
- In stories: `Table` and `Simple Table` to `Legacy Table/Table` and `Legacy Table/Simple Table`, respectively

Because the new Table component is to be created in the now-existing `src/table` dir, latter should be renamed in a separate commit beforehand. Otherwise `git` would think that the new component is `legacy-table`, not `table`.


Hypnosphi:
Please reflect it in `CHANGELOG.md`

Hypnosphi (approved):

andrey-skl (approved):

Hypnosphi:
@aleksei-berezkin please also rename `table.css` to `legacy-table.css`, otherwise classnames will clash in the built version when you add the new `table/table.css`. They are currently generated from just the file name: https://github.com/JetBrains/ring-ui/blob/c31cae29972890af1738dcb5f48c9c80f3678542/packages/rollup-css-plugin/css-plugin.js#L86

PR #9307, opened 2026-06-09
RG-2785 Fixed parsing to invalid date and time without date input

aleksei-berezkin:
This was a regression after the Date Picker rework

### DIFF
--- a/src/date-picker/date-picker.tsx
+++ b/src/date-picker/date-picker.tsx
@@ -210,19 +210,18 @@ export default class DatePicker extends PureComponent<DatePickerProps> {
   };
 
   parse = memoize((date: Date | number | string | null | undefined) => {
-    const {parseDateInput} = this.props;
+    let normalizedDate;
 
     if (date instanceof Date) {
-      return isValid(date) ? date : null;
+      normalizedDate = date;
+    } else if (typeof date === 'number') {
+      normalizedDate = new Date(date);
+    } else {
+      const {parseDateInput} = this.props;
+      normalizedDate = parseDateInput(date);
     }
 
-    if (typeof date === 'number') {
-      const dateFromNumber = new Date(date);
-      return isValid(dateFromNumber) ? dateFromNumber : null;
-    }
-
-    const parsedDate = parseDateInput(date);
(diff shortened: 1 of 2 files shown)


src/date-picker/date-picker.tsx
```diff
@@ -213,7 +219,11 @@ export default class DatePicker extends PureComponent<DatePickerProps> {
       return new Date(date);
     }
 
-    return parseDateInput(date);
+    const parsedDate = parseDateInput(date);
+    if (parsedDate && isValid(parsedDate)) {
+      return parsedDate;
+    }
+    return null;
```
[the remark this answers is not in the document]
aleksei-berezkin: Fixed

src/date-picker/date-picker.tsx
```diff
@@ -207,13 +213,16 @@ export default class DatePicker extends PureComponent<DatePickerProps> {
...
+
     if (typeof date === 'number') {
-      return new Date(date);
+      const dateFromNumber = new Date(date);
+      return isValid(dateFromNumber) ? dateFromNumber : null;
     }
 
-    return parseDateInput(date);
+    const parsedDate = parseDateInput(date);
+    return parsedDate && isValid(parsedDate) ? parsedDate : null;
```
vkrol: I won't insist, but I would remove the duplicate `isValid` calls so we don't forget about it later if we need to add another branch here:
```ts
parse = memoize((date: Date | number | string | null | undefined) => {
  let normalizedDate;

  if (date instanceof Date) {
    normalizedDate = date;
  } else if (typeof date === 'number') {
    normalizedDate = new Date(date);
  } else {
    const {parseDateInput} = this.props;
    normalizedDate = parseDateInput(date);
  }

  return normalizedDate && isValid(normalizedDate) ? normalizedDate : null;
});
```

vkrol (approved):

andrey-skl (approved):

src/date-picker/date-picker.tsx
aleksei-berezkin: Thx, replaced the implementation with the suggested snippet

vkrol (approved):

PR #9318, opened 2026-06-19
Fix EditableHeading multiline sizing on edit when autofcus is false

vkrol:
CleanShot 2026-06-19 at 14 46 55@2x

### DIFF
--- a/src/editable-heading/editable-heading.tsx
+++ b/src/editable-heading/editable-heading.tsx
@@ -1,4 +1,4 @@
-import {type InputHTMLAttributes, useEffect} from 'react';
+import {type InputHTMLAttributes, useEffect, useLayoutEffect} from 'react';
 import * as React from 'react';
 import classNames from 'classnames';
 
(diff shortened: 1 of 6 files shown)


src/editable-heading/editable-heading.tsx
```diff
@@ -125,30 +125,33 @@ export const EditableHeading = (props: EditableHeadingProps) => {
     inputClassName,
   );
 
-  const stretch = (el: HTMLElement | null | undefined) => {
+  const stretch = React.useCallback((el: HTMLElement | null | undefined) => {
```
aleksei-berezkin: `stretch()` has only one usage, so it can be inlined into `checkValue()`

aleksei-berezkin (approved):

src/editable-heading/editable-heading.tsx
vkrol: Thanks! Done https://github.com/JetBrains/ring-ui/pull/9318/commits/1d123f94e201ffdf4e18b10eccc67f51f1c55242

PR #9340, opened 2026-07-07
JT-93843 coordinate token refresh across tabs and retry stale-token 401

andrey-skl:
## Problem

[JT-93843](https://youtrack.jetbrains.com/issue/JT-93843): users with several YouTrack tabs open intermittently get logged out (401) and have to reload — even on an active machine. The earlier retry fix (#9203) only covered transient network failures in the background flow and didn't help.

Console logs from affected users show the 401 surfacing from `_detectUserChange` (`RingUI Auth: Init failure ... at _detectUserChange`) — a path #9203 never touched.

## Root cause

1. **Cross-tab refresh race (thundering herd).** When tokens expire around the same time, every open tab races Hub with its own silent iframe refresh. The resulting token churn hands some tabs an already-superseded token, which then 401s. `forceTokenUpdate` only deduped within a single tab. This also explains why shortening the token lifetime anecdotally reduced the failure rate.
2. **No retry on the passive path.** `_detectUserChange` (fired by a cross-tab token-storage change) called `getUser` with the handed token and went straight to the auth dialog on 401.

## Fix

**Cross-tab coordination** (`auth-core.ts`):
- Silent refresh is serialized across tabs via the Web Locks API (`ring-ui-token-refresh-<clientId>`). The lock holder refreshes; tabs that acquire the lock afterwards reuse the token the holder just wrote — but only when the stored token actually **changed** while waiting. An unchanged (server-rejected yet locally valid) token still forces a real refresh, so the reuse fast-path can't defeat a 401 retry.
- Callers that know which token the server rejected (`http.request` retry, `_detectUserChange`) pass it explicitly as the comparison baseline. The baseline for plain `forceTokenUpdate()` is captured **before** the backend status check, so a sibling's refresh during that network wait reads as "changed".
- The lock callback never shows dialogs, never redirects, and never rejects — it resolves to `{token} | {error}`, so the lock is always released promptly. Interactive failure handling (`_handleTokenRefreshFailure`) runs after release.
- Lock acquisition is bounded by a 15s `AbortSignal.timeout` (a timer-throttled background tab could otherwise hold waiters for minutes) and falls back to an unsynchronized refresh, as it does where Web Locks is unavailable.

**Passive 401 retry** (`auth-core.ts`):
- `_detectUserChange` → `_getUserWithTokenRefresh`: on 401, refresh once (rejected token as baseline) and retry `getUser`. A failed refresh rethrows the original 401 instead of triggering the interactive dialog/redirect fallback — preserving old behavior for consumers without embedded login (Hub, TeamCity).

**HTTP retry baseline** (`http.ts`):
- `HTTP.request()`'s refresh-and-retry now passes the rejected token to `forceTokenUpdate(failedToken?)` (backwards-compatible optional param), so a fresh token another tab already wrote is reused instead of minting yet another one.

## Testing

- 7 new unit tests: sibling-token reuse (via lock and during backend check), unchanged-token forces refresh, explicit rejected-token baseline, lock-timeout fallback, passive 401 retry, failed-refresh rethrow (no dialog/redirect), non-401 propagation. Key tests verified to fail against the pre-fix code.
- Verified live in Storybook (`Utilities/Auth → basic`) against hub.jetbrains.com with `navigator.locks` instrumentation: lock acquired with correct name + AbortSignal; sibling-written token reused with no redundant refresh; permanently-held lock times out after 15s and falls back to a working unsynchronized refresh; init recovers silently from a server-rejected stored token.

### DIFF
--- a/rfc/JT-93843-Token-Refresh-Coordination.md
+++ b/rfc/JT-93843-Token-Refresh-Coordination.md
@@ -25,7 +25,7 @@ Ring UI coordinates silent token refresh with the Web Locks API and reuses a fre
 
 When no rejected token is known, `forceTokenUpdate()` snapshots the stored token before the backend status check. This keeps a sibling refresh that happens during the backend check visible as a changed token.
 
-Silent refresh runs under a Web Lock named by `clientId`. The lock callback only returns `{token}` or `{error}`. Dialog and redirect fallback are handled after the lock is released.
+Silent refresh runs under a Web Lock named by `clientId`. The lock callback normalizes refresh results into either a token result or an error result. Dialog and redirect fallback are handled after the lock is released.
 
 Lock acquisition is bounded by `TOKEN_REFRESH_LOCK_TIMEOUT`. If the Web Locks API is unavailable, lock acquisition times out, or lock acquisition rejects, the tab falls back to the previous unsynchronized refresh behavior.
 
--- a/rfc/JT-93843-Token-Refresh-Coordination.md
+++ b/rfc/JT-93843-Token-Refresh-Coordination.md
@@ -1,0 +1,42 @@
+# JT-93843 Token Refresh Coordination
+
+## Summary
+
+Multiple YouTrack tabs can refresh an access token at the same time after inactivity, sleep, or network recovery. Each silent iframe refresh can supersede the token written by another tab. A tab that receives an already-superseded token then gets a 401 and may show an auth dialog or log the user out until the page is refreshed.
+
+Ring UI coordinates silent token refresh with the Web Locks API and reuses a fresh token written by another tab instead of minting another one.
+
+## Goals
+
+- Serialize silent token refresh across tabs and same-page Ring UI clients that share the same `clientId`.
+- Reuse a locally valid stored token when it differs from the token that failed or from the pre-refresh storage snapshot.
+- Still force a real refresh when the stored token is unchanged, because that token can be locally valid yet server-rejected.
+- Keep interactive dialog and redirect handling outside the Web Lock callback.
+- Retry passive `_detectUserChange` user loading once after a 401 without invoking the interactive refresh fallback.
+
+## Non-goals
+
+- Track all rejected tokens.
+- Add a custom cross-tab locking implementation for browsers without Web Locks.
+
+## Design
+
+`HTTP.request()` keeps the token used for the failed request and passes it to `forceTokenUpdate(failedToken)`. That token is the comparison baseline. If another tab has already written a different locally valid token, Ring UI reuses it. If storage still contains the same token, Ring UI refreshes through Hub.
+
+When no rejected token is known, `forceTokenUpdate()` snapshots the stored token before the backend status check. This keeps a sibling refresh that happens during the backend check visible as a changed token.
+
+Silent refresh runs under a Web Lock named by `clientId`. The lock callback only returns `{token}` or `{error}`. Dialog and redirect fallback are handled after the lock is released.
+
+Lock acquisition is bounded by `TOKEN_REFRESH_LOCK_TIMEOUT`. If the Web Locks API is unavailable, lock acquisition times out, or lock acquisition rejects, the tab falls back to the previous unsynchronized refresh behavior.
+
+`_detectUserChange()` uses the access token from the storage event as the rejected-token baseline when `getUser()` returns 401. On refresh failure, it rethrows the original 401 so the passive storage-event path does not run the interactive force-refresh fallback.
+
+## Rejected Alternatives
+
+### Store the last issued token in `Auth`
+
+One mutable `lastIssuedToken` is not per request. A later request in the same tab can overwrite it before an earlier request handles its 401, causing the earlier request to compare against the fresh stored token and mint another token.
+
+### Store all rejected token hashes
+
+This adds state and cleanup rules without removing the need to know which token failed. The per-request rejected token is already available in `HTTP.request()`.
--- a/src/auth/auth-core.ts
+++ b/src/auth/auth-core.ts
@@ -205,6 +205,10 @@ class Auth implements HTTPAuth {
   private _postponed = false;
   private _backendCheckPromise: Promise<void> | null = null;
   private _forceTokenUpdatePromise: Promise<string | null> | null = null;
+  // The last token this instance handed out. Used as the refresh baseline: if
+  // the stored token differs from it (e.g. another tab refreshed), the stored
+  // one is reused instead of minting yet another token (JT-93843).
+  private _lastIssuedToken: string | null = null;
   private _authDialogService: typeof AuthDialogService | undefined = undefined;
   _domainStorage: AuthStorage<UserChange>;
   user: AuthUser | null = null;
@@ -531,7 +535,9 @@ class Auth implements HTTPAuth {
       if (Auth.storageIsUnavailable) {
         return null; // Forever guest if storage is unavailable
       }
-      return (await this._tokenValidator?.validateTokenLocally()) ?? null;
+      const token = (await this._tokenValidator?.validateTokenLocally()) ?? null;
+      this._lastIssuedToken = token;
+      return token;
     } catch (e) {
       return this.forceTokenUpdate();
     }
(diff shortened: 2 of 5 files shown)


src/auth/auth-core.ts
```diff
@@ -541,19 +547,30 @@ class Auth implements HTTPAuth {
...
    * the Hub session is re-established.
    *
+   * @param failedToken the token the caller knows was rejected by the server,
+   * if any. Used as the reuse baseline: a stored token that differs from it
+   * (e.g. written by another tab) is returned as is instead of minting yet
+   * another one (JT-93843).
    * @return {Promise.<string | null>}
    */
-  forceTokenUpdate(): Promise<string | null> {
+  forceTokenUpdate(failedToken?: string | null): Promise<string | null> {
```
PavelReutov, line 541: I'm a little bit concerned that we are changing the API for this class. And the failed token is actually coming from the same auth instance. The code looks like (with some simplification):

```
const token = auth.requestToken()
try{
...
}
catch(){
auth.forceTokenUpdate(token)
}
```

maybe we can encapsulate this logic inside the `auth` class? For example store like `lastToken` inside `auth`.

PavelReutov (approved):

src/auth/auth-core.ts
```diff
@@ -565,19 +592,78 @@ class Auth implements HTTPAuth {
...
+    }
+    const lockName = `ring-ui-token-refresh-${this.config.clientId}`;
+    try {
+      return await navigator.locks.request(lockName, {signal: AbortSignal.timeout(TOKEN_REFRESH_LOCK_TIMEOUT)}, () =>
+        this._tryRefreshToken(previousToken),
+      );
+    } catch {
+      // Lock acquisition failed or timed out — the callback itself never
+      // rejects. Refresh unsynchronized rather than stall this tab.
+      return this._tryRefreshToken(previousToken);
```
aleksei-berezkin: To avoid guessing whether the callback can reject, I'd handle this case with 

```js
AbortSignal.timeout(...).onAbort(() => /* handle here */ )
```

src/auth/auth-core.ts
```diff
@@ -565,19 +592,78 @@ class Auth implements HTTPAuth {
...
+    // reuse it instead of triggering a redundant refresh (JT-93843). Only skip
+    // the refresh when the stored token actually changed — otherwise the token
+    // we hold is still the (server-rejected) one and must be replaced.
+    let storedToken: string | null = null;
+    try {
+      storedToken = (await this._tokenValidator?.validateTokenLocally()) ?? null;
+    } catch {
+      // No valid local token — fall through to refresh.
+    }
+    if (storedToken != null && storedToken !== previousToken) {
```
aleksei-berezkin, line 593: This resembles the "compare-and-set" pattern, so maybe use it in the name? I.e. `compareAndRefresh`, or, if you're not sure if a reader is familiar with the pattern, sth like `refreshIfNotChanged`

src/auth/auth-core.ts
```diff
@@ -565,19 +592,78 @@ class Auth implements HTTPAuth {
...
+   * The lock callback never shows dialogs or redirects and never rejects — it
+   * resolves to `{token}` or `{error}` — so the lock is always released
+   * promptly and interactive failure handling runs outside the lock.
+   *
+   * Falls back to an unsynchronized refresh where the Web Locks API is
+   * unavailable (older browsers, insecure contexts, tests) or when the lock
+   * cannot be acquired within {@link TOKEN_REFRESH_LOCK_TIMEOUT} (e.g. held by
+   * a timer-throttled background tab).
+   */
+  private async _tryRefreshTokenAcrossTabs(previousToken: string | null): Promise<TokenRefreshAttempt> {
```
aleksei-berezkin: This code will help fixing the races not only between tabs but within the same tabs if i.e. you have several frontend apps within the same page, and each uses Ring UI. So maybe `refreshUnderWebLock`

aleksei-berezkin (changes requested):
The comments verbosity stands out for me, to be honest. Maybe somewhat sharper names would allow contracting them.

src/auth/auth-core.ts
andrey-skl: I have applied it in the dedicated [commit](https://github.com/JetBrains/ring-ui/pull/9340/commits/e7db1053dca68a0a4bf3304e18825512c645d795), but verifying it with LLMs I have a concern:

- **P1** [src/auth/auth-core.ts:582](/Users/skl/Documents/github/ring-ui/ring-ui-master/src/auth/auth-core.ts:582): `_lastIssuedToken` is instance-global, so it is not reliably “the token this failed request used.” If request A gets `old-token`, another tab writes `fresh-token`, then request B in the same tab calls `requestToken()` before A handles its 401, line 539 overwrites `_lastIssuedToken` with `fresh-token`. A then calls `forceTokenUpdate()`, uses `fresh-token` as the baseline, sees storage unchanged, and mints yet another token. That preserves the token-churn race the previous explicit `forceTokenUpdate(failedToken)` version avoided.

So this approach is race-condition prone, and maybe previous argument-based forceTokenUpdate(failedToken) is better. What is your opinion?

PavelReutov: Thanks, for the explanation, I see. Your case is totally valid. But I still have some concerns with this approach.
But I'm not entirely sure that the case when we store `old-token` outside the class still solves the problem. Because we still can get two calls of the `forceTokenUpdate` and one will end up faster and we will overwrite the fresh token because we scheduled the update in the different task. But this is a hypothetical scenario.
Maybe we should store all expired tokens of at least their hash? And ignore the request to update the token if it's already known that it's invalid

src/auth/auth-core.ts
```diff
@@ -541,19 +553,34 @@ class Auth implements HTTPAuth {
...
+      });
     return this._forceTokenUpdatePromise;
   }
 
   private async _doForceTokenUpdate(): Promise<string | null> {
+    // The refresh baseline: the last token this instance handed out — the one
+    // the caller deems bad. When nothing was issued yet, fall back to the
+    // stored token, captured before the backend check so a sibling tab's
+    // refresh during that network wait reads as "changed" (JT-93843).
+    const previousToken = this._lastIssuedToken ?? (await this._storage?.getToken())?.accessToken ?? null;
```
PavelReutov: Maybe we should acquire token lock before reading the storage? or at least check if there is a pending lock. Because when we are awaiting the storage in theory the other thread can update the token and we will reset the token one more time.

@claude we need your opinion on this

src/auth/auth-core.ts
andrey-skl: Thanks! After looking into this, I think the current try/catch is the right shape here:

AbortSignal.timeout() does not expose `onAbort()`, and `navigator.locks.request()` already rejects when the acquisition signal aborts. The lock callback itself is intentionally normalized to {token} / {error}, so the catch handles lock acquisition failure rather than refresh failure.

Would you agree?

src/auth/auth-core.ts
andrey-skl: Applied `_refreshTokenUnderWebLock`, thanks!

src/auth/auth-core.ts
andrey-skl: Agree, renamed to _refreshTokenOrReuseChanged

andrey-skl:
> The comments verbosity stands out for me, to be honest. Maybe somewhat sharper names would allow contracting them.

@aleksei-berezkin thanks! I have decided to extract the narrative from the comments to RFC, applied in https://github.com/JetBrains/ring-ui/pull/9340/commits/b1cc7b6ca7b8e84265e4527c2ab4acd045a62145

src/auth/auth-core.ts
andrey-skl: Reverted as storing last token is clearly not reliable https://github.com/JetBrains/ring-ui/pull/9340/commits/2591ba152edcb38dbe62ea0d5d0d1ffa16626bce

andrey-skl:
> Maybe we should store all expired tokens of at least their hash?

@PavelReutov honestly I think it would over overcomplicate things, without proven benefits. Requires careful management of the storage, cleanup, etc.

src/auth/auth-core.ts
andrey-skl: I don’t think acquiring the Web Lock before the initial storage snapshot helps. When the caller has a rejected token, that token is the correct baseline and no storage read is needed for correctness. When the caller does not have one, reading storage before backend check / lock wait is intentional: if another tab updates storage while we wait, the later compare sees a changed token and reuses it. Checking for a pending lock would duplicate what `navigator.locks.request(..., {signal})` already does.

rfc/JT-93843-Token-Refresh-Coordination.md
```diff
@@ -0,0 +1,42 @@
...
+- Track all rejected tokens.
+- Add a custom cross-tab locking implementation for browsers without Web Locks.
+
+## Design
+
+`HTTP.request()` keeps the token used for the failed request and passes it to `forceTokenUpdate(failedToken)`. That token is the comparison baseline. If another tab has already written a different locally valid token, Ring UI reuses it. If storage still contains the same token, Ring UI refreshes through Hub.
+
+When no rejected token is known, `forceTokenUpdate()` snapshots the stored token before the backend status check. This keeps a sibling refresh that happens during the backend check visible as a changed token.
+
+Silent refresh runs under a Web Lock named by `clientId`. The lock callback only returns `{token}` or `{error}`. Dialog and redirect fallback are handled after the lock is released.
```
aleksei-berezkin: You probably mean `{token: string}` and `{error: string}` here

aleksei-berezkin (approved):

rfc/JT-93843-Token-Refresh-Coordination.md
andrey-skl: Agree, fixed in https://github.com/JetBrains/ring-ui/pull/9340/commits/4fa3655169e52808114c659ba6d8d6786d9dca40

PR #9343, opened 2026-07-10
RG-2542 Table: item reorder with refined API

aleksei-berezkin:
Rows can now be reordered by dragging or with arrow keys. The built-in drag frame and insertion indicator work out of the box, with the full drag lifecycle exposed via `onUserDrag()` for custom visuals. 

Reordering works as expected with row virtualization and custom multi-row renderers.

Additionally, manual `useCallback`/`useMemo` wrappers were removed from several internal hooks, delegating memoization to the React Compiler.

*This change concludes the required set of Table features. The only feature left is column resizing, which I would like to postpone so far (but I'm not refusing to implement it, just somewhat later). After we merge this PR, it would be nice to start the 8.0 and the Table adoption.*


src/table/internal/reorder-animation-context.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  function expectReorder(reorderSpec: ReorderSpec) {
+    const isColumn = reorderSpec.direction === 'columns';
+    if ((isColumn && noColumnReorderAnimation) || (!isColumn && noItemReorderAnimation)) return;
+
+    const timerId = window.setTimeout(() => {
+      pendingReorderRef.current = null;
+    }, pendingReorderTimeoutMs);
+    pendingReorderRef.current = {...reorderSpec, timerId, columns, data};
+  }
```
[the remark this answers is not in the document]
aleksei-berezkin: Fixed

src/table/table.tsx
```diff
@@ -214,11 +231,20 @@ import styles from './table.css';
  * set the `noItemVirtualization` prop to `true`, otherwise it will also try
  * to control the virtualization, possibly reporting incorrect item height.
  *
- * ### Column reorder animation
+ * ### Item reorder
  *
- * Default-rendered rows highlight the column that was just reordered. To apply
- * the same animation to your custom-rendered rows, use `ColumnAnimationContext`
- * (from `table/table-const`) to get information about the currently animated column.
+ * If `Table.onItemReorder` is set and your item spans multiple rows, call
```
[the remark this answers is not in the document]
aleksei-berezkin: Fixed

src/table/table-primitives.tsx
```diff
@@ -36,3 +38,51 @@ export function TableCell(props: ComponentPropsWithRef<'td'>) {
...
+
+/**
+ * - `'pointerdown'` means the user has pressed the mouse button or touched the handle.
+ * - `number` means the distance in pixels the user has dragged the handle along
+ *   the Y axis.
+ * - `'cancelled'` means the reorder was aborted — e.g. the user pressed Escape,
+ *   released the pointer outside the browser, or dropped the item at its original
+ *   position. Use this phase to play a cancellation animation in your custom renderer.
+ *   This state lasts 600 ms (matching the built-in cancel animation), after which
+ *   `undefined` is sent.
```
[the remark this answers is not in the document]
aleksei-berezkin: The client is not required to use the full 600 ms window. They may run shorter animation, and cleanup after an additional delay. That said if some client needs a different animation window, the API will be refined. For the moment it's unclear if it's needed at all.

src/table/internal/reorder-handle.tsx
```diff
@@ -0,0 +1,466 @@
...
+  }
+
+  function handlePointerUp(e: PointerEvent<HTMLButtonElement>) {
+    onPointerUp?.(e);
+    if (e.defaultPrevented) return;
+
+    if (activeDragRef.current?.state !== 'is-dragging') return;
+
+    const {clientX, clientY} = e;
+    const insertionPoint = getClosestInsertionPointLocal(clientX, clientY);
```
andrey-skl, line 379: AI-assisted finding:

If `canReorderItem` rejects every insertion point, `getClosestInsertionPointLocal()` returns `itemIndex: -1`, so this calls `onItemReorder(fromIndex, -1)`. Please treat `itemIndex === -1` as a cancelled drag before calculating `insertionIndex`.

src/table/internal/reorder-animation-context.ts
```diff
@@ -0,0 +1,121 @@
...
+    const {direction, data: pendingData, columns: pendingColumns} = pendingReorder;
+    const isColumn = direction === 'columns';
+    if ((isColumn && columns === pendingColumns) || (!isColumn && data === pendingData)) return;
+
+    pendingReorderRef.current = null;
+
+    const {fromIndex, insertionIndex} = pendingReorder;
+
+    // Moving forward shifts the array by one, hence -1
+    const index = fromIndex < insertionIndex ? insertionIndex - 1 : insertionIndex;
```
andrey-skl, line 90: Didn't you in general consider using stable keys instead of the index? In case of any changes in the data, like async filtering, it would be a problem

aleksei-berezkin: Thx, good catch. Are you only talking here about column/item reorder animation, or do you have concerns regarding any other use cases? So far I don't see any other places where this may be a problem because most of the code is synchronous:

```tsx
// client.tsx
<Table
  data={args.data}
  columns={columns}
  renderItem={(_item, index) => (
    // synchronous
    <DefaultItemRenderer index={index} />
  )}
  onItemReorder={(fromIndex, insertionIndex, items) =>
    // synchronous
    reorderItems(items, fromIndex, insertionIndex, setData)
  }
  onColumnDelete={(columnIndex, columns) =>
    // synchronous
    setColumns(columns.filter((_, i) => i !== columnIndex))
  }
/>
```

In the last 2 cases, even if items and columns need to be refetched before update, the client has an access both to `data` and `columns` arrays as last params in the handlers, and can easily get keys or any other stable field before asynchronously updating data. Note: backend items/columns `id` doesn't necessarily equals `TableProps.getKey()`, that's why, I believe, just providing an index here is the less guessing what the client needs.

So the animation after the expected reorder seems to me the only affected feature. Or do you have anything else on your mind?

andrey-skl: I can imagine developers mixing these concepts in that way:

```ts
const theirData = ... // their data, maybe even mutable

...
onItemReorder={(fromIndex, insertionIndex, items) =>
    const reordering item = theirData[fromIndex];
    
    ... // If by any chance `theirData` was updated in the meantime, it could cause desync
  }
  
  ```
  
  So maybe we should also pass the `draggingItem` down to the callback to make it both easy and safe to access it?

Comment (aleksei-berezkin) [reply src/table/internal/reorder-animation-context.ts:90]: Okay so do you mean of redesigning the method like this:

```ts
interface TableProps<T> {
  // ...
  onItemReorder(
    draggedItem: T,          // A row user actually moved by dragging
    fromIndex: number,       // An index of draggedItem
    insertionIndex: number,  // Where to insert
    items: T[]               // All current items
  )
}
```

And the client:

```ts
const theirData = ... // their data, maybe even mutable

...
onItemReorder={(draggedItem, fromIndex, insertionIndex, items) =>
  // theirData[fromIndex] is not needed anymore -- it's passed as draggedItem

  // ... But the problem with theirData/items desync remains
  // If by any chance `theirData` was updated in the meantime, it could cause desync
}
```

So far I don't see honestly how can we help the client avoid the resync issues. Do I miss sth? Could you pls provide some more pseudocode which helps in this situation?

andrey-skl: Well maybe I am could not proof the way of data to get desynchronized, but adding `draggedItem` to the arguments makes it more flexible, at least user would have a direct way to read the dragged item

Can we really have it?

Hypnosphi (approved):

src/table/internal/reorder-handle.tsx
aleksei-berezkin: Instead of `-1` I'm returning now `undefined` and checking it on a call site.

src/table/internal/reorder-animation-context.ts
aleksei-berezkin: I added `item` and `column` as a first argument in all methods related to column and item delete and reorder.

andrey-skl (approved):

PR #9359, opened 2026-07-31
RG-2791 Alias imported `button` value and guard built CSS against `@value` ident collisions

Hypnosphi:
## Problem

The CSS-modules `@value` import replacement (icss-utils `replaceValueSymbols`, run by `postcss-modules` inside `@jetbrains/rollup-css-plugin`) rewrites the imported identifier as a bare token **everywhere** in the file — selectors, at-rule params, declaration values — not only in `.class` positions.

#9353 fixed the `active` collision (`:active` → invalid `:ring-button-active`, `(-ms-high-contrast: active)` → invalid media query). But the published package still ships the same class of corruption for the `button` value: element selectors like `.common > button:first-child` and `.buttonToolbar > button` are emitted as bare `ring-button-button` — an unknown element selector, so those rules (button-group corner radii, toolbar margins for plain `<button>`s) are dead in every browser. Verified against the published `@jetbrains/ring-ui-built` `components/style.css`.

## Fix

- Alias `button as buttonClass` in `button-group.css` and `button-toolbar.css` (same approach as #9353) and rename the `.button` class usages; bare `button` element selectors are untouched and now safe. No TSX code references `styles.button` from these files.

## Regression guard

- New `scripts/check-built-css.js`, run in `postbuild`: parses `dist/style.css` and fails the build on any scoped class name leaking outside class-selector position — pseudo-classes (`:ring-*`), bare element selectors, attribute names/values, functional pseudo args, at-rule params (`@media`, `@supports`, …), declaration values (with `var(--ring-*)`, declared `@keyframes` animation names, and `.ring-*` class selectors exempted where legitimate) — plus a positive check that the button-group `:active` rules survived.
- Unit tests for the checker in `src/global/check-built-css.test.ts` (19 cases covering each corruption context and the legitimate `ring-*` usages).

Verified locally: `npm run build` passes the new check, and the built `style.css` contains intact `.ring-button-group-common > button` selectors, zero `:ring-button-active`, and correct `(-ms-high-contrast:active)` media queries.

### DIFF
--- a/scripts/check-built-css.js
+++ b/scripts/check-built-css.js
@@ -1,0 +1,99 @@
+/*
+ * RG-2791: CSS-modules `@value` import replacement (icss-utils replaceValueSymbols) renames
+ * the imported identifier everywhere in the file, not only in `.class` positions. When an
+ * imported value name collides with another ident (`:active` pseudo-class, `button` element
+ * selector, `(-ms-high-contrast: active)` media feature value), the built CSS ends up with
+ * invalid selectors like `:ring-button-active` or bare `ring-button-button`, and browsers
+ * drop those rules entirely. Guard the built stylesheet against that.
+ */
+/* eslint-disable no-console */
+const fs = require('fs');
+const path = require('path');
+const postcss = require('postcss');
+
+// A scoped class name leaking outside class-selector position. In selectors a `ring-*` ident is
+// only legitimate right after `.` (class selector); in declaration values only inside a longer
+// ident such as `var(--ring-*)` or, for animation properties, as a name actually declared by a
+// scoped `@keyframes` in the same stylesheet. Everything else — `:ring-button-active` (was `:active`), bare
+// `> ring-button-button` (was `> button`), `:is(ring-x)`, `[ring-x]`, `[type=ring-x]`,
+// `12px/ring-x`, `content:'ring-x'` — is corruption left by the CSS-modules @value replacement.
+// ponytail: an intentional ring-prefixed attribute/string value (e.g. `[class~='ring-x']`) would
+// be a false positive — no such CSS exists; add an allowlist here if one ever appears
+const LEAK = /(?<![\w.-])ring-[\w-]+/;
+// In declaration values a `.` proves nothing (a leaked `content: ".ring-button-button"` would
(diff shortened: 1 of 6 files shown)


scripts/check-built-css.js
```diff
@@ -0,0 +1,99 @@
...
+ * the imported identifier everywhere in the file, not only in `.class` positions. When an
+ * imported value name collides with another ident (`:active` pseudo-class, `button` element
+ * selector, `(-ms-high-contrast: active)` media feature value), the built CSS ends up with
+ * invalid selectors like `:ring-button-active` or bare `ring-button-button`, and browsers
+ * drop those rules entirely. Guard the built stylesheet against that.
+ */
+/* eslint-disable no-console */
+const fs = require('fs');
+const path = require('path');
+const postcss = require('postcss');
```
aleksei-berezkin: `postcss` supports ESM imports

scripts/check-built-css.mjs
```diff
@@ -0,0 +1,99 @@
...
+
+  // Positive check: button-group :active rules survived the build intact
+  let hasButtonGroupActive = false;
+  root.walkRules(rule => {
+    if (rule.selector.includes('.ring-button-group-') && rule.selector.includes(':active')) {
+      hasButtonGroupActive = true;
+    }
+  });
+  if (!hasButtonGroupActive) {
+    errors.push('Expected at least one button-group rule with the :active pseudo-class');
```
aleksei-berezkin, line 85: Pls consider adding a comment (or an error msg) that if `button-croup.css` is reworked, this check may start failing, which may be expected, and that that some other positive check should be added in this case

aleksei-berezkin (approved):

andrey-skl (approved):

scripts/check-built-css.mjs
Hypnosphi: added: https://github.com/JetBrains/ring-ui/pull/9359/changes/1fabadfad603abe9329ba19aeed42c0e34b4a37a#diff-95e868c12159d93615e4976ed3e347cb8600ad95e12003b35f011961d64a4f2eR75

scripts/check-built-css.js
Hypnosphi: switched to .mjs

PR #9360, opened 2026-07-31
Add keepMounted and headingLevel props to Collapse/CollapsibleGroup

Hypnosphi:
## What

- **`keepMounted`** (opt-in, default off) on `Collapse` and `CollapsibleGroup`: keeps collapsed content mounted — hidden via `visibility: hidden` + `inert` — instead of unmounting it, so local state, subscriptions, and iframes survive collapse cycles. Documented limitations: portaled content (e.g. `Popup`) is not hidden, and hidden form controls still participate in constraint validation.
- **`headingLevel?: 2–6`** on `CollapsibleGroup`: wraps the disclosure header in an `<h2>`–`<h6>` (standard `<hN><button aria-expanded>` accordion pattern) to preserve the document outline.

## Behavior changes / fixes along the way

- `CollapseContent` is now `inert` whenever the panel is collapsed, matching `aria-expanded`: content animating shut can't be tabbed into, and a `minHeight` preview is a visual-only teaser (see the `minHeight` JSDoc for the AT trade-off).
- Collapse finalization only reacts to the container's **own `height` transitionend** (descendant transitions used to end it early), with a once-per-collapse fallback timer derived from the committed `--duration` for suppressed/cancelled transitions (`transition: none`, reduced motion).
- `disableAnimation` no longer unmounts a positive-`minHeight` preview, and stays in sync when toggled while collapsed.
- `keepMounted` is optional in `CollapseContext`, so externally constructed `Provider` values keep compiling.
- `inert` is serialized cross-version (React 19 boolean vs React 18 empty-string form) and rendered in JSX so SSR markup is protected before hydration. TODO markers note that the React 18 branch should be dropped in develop-8.0.
- The deprecated `expand/collapsible-group.css` alias forwards the new `heading` class, guarded by a token-parity + composes test.

## CI

- New **React 18 runtime compatibility** TeamCity build (`.teamcity/src/tests/React18Compat.kt`), wired as a snapshot dependency of `All checks`. ⚠️ The DSL is untested locally — please sanity-check it compiles on the TeamCity side.

## Tests

25 unit tests across `src/collapse` / `src/collapsible-group`, covering: state preservation across a settled collapse cycle, immediate `inert` on collapse, transitionend target/property filtering, fallback-timer behavior under resize churn / mid-collapse `duration` changes / stale-height starts, SSR markup, `minHeight` previews, and heading semantics.

### DIFF
--- a/src/collapse/collapse-content.tsx
+++ b/src/collapse/collapse-content.tsx
@@ -83,6 +83,7 @@ export const CollapseContent: React.FC<PropsWithChildren<Props>> = ({
     // Armed once per collapse toggle from the --duration committed to the DOM — the value
     // the running transition actually uses — so it can neither undercut a transition started
     // from a stale height nor be restarted by content resizes or duration changes mid-collapse
+    // TODO merge with global/parse-css-duration when this lands in develop-8.0
     const cssDuration = parseFloat(container?.style.getPropertyValue('--duration') || '') || 0;
     const fallbackTimeout = window.setTimeout(finalizeCollapse, cssDuration + HIDE_FALLBACK_EXTRA_DELAY);
 
@@ -92,6 +93,9 @@ export const CollapseContent: React.FC<PropsWithChildren<Props>> = ({
     };
   }, [collapsed, initialContentHeight]);
 
+  // render-phase state adjustments (not effects): the React Compiler lint forbids
+  // setState-in-effect, and https://react.dev/learn/you-might-not-need-an-effect
+  // documents this pattern for resetting state when props change
   if (!collapsed && shouldHideContent) {
     setShouldHideContent(false);
   }
--- a/src/collapse/collapse-content.tsx
+++ b/src/collapse/collapse-content.tsx
@@ -13,8 +13,29 @@ const DURATION_FACTOR = 0.5;
 const DEFAULT_HEIGHT = 0;
 const VISIBLE = 1;
 const HIDDEN = 0;
+// margin for the hide fallback timer, so it fires only if transitionend never did
+const HIDE_FALLBACK_EXTRA_DELAY = 100;
+
+// React 18 renders unknown attributes from strings only, while React 19 treats `inert`
+// as a real boolean and removes the attribute when it gets ''
+// TODO drop the React 18 branch in develop-8.0 — Ring UI 8.0 supports React 19 only
+const REACT_MAJOR_WITH_INERT_SUPPORT = 19;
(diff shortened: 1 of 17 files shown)


src/collapse/collapse.tsx
```diff
@@ -8,6 +8,15 @@ interface Props {
...
   disableAnimation?: boolean;
+  /**
+   * Keep children mounted while collapsed (hidden via CSS)
+   * instead of unmounting them, preserving their state.
+   * Note: hidden form controls still participate in form validation
+   * and submission — disable them while collapsed if needed.
+   * Content rendered through portals (e.g. Popup) escapes the hidden
+   * wrapper and is not hidden — close overlays on collapse.
+   */
+  keepMounted?: boolean;
```
aleksei-berezkin, line 19: Would be nice documenting the exact way of hiding it with CSS (`visibility: 'hidden'`)

src/collapse/collapse-content.tsx
```diff
@@ -41,24 +62,45 @@ export const CollapseContent: React.FC<PropsWithChildren<Props>> = ({
...
+      }
+    }
+
     container?.addEventListener('transitionend', onTransitionEnd);
+    // fallback for suppressed or cancelled transitions (e.g. `transition: none` overrides),
+    // which emit no transitionend and would otherwise leave the content interactive forever.
+    // Armed once per collapse toggle from the --duration committed to the DOM — the value
+    // the running transition actually uses — so it can neither undercut a transition started
+    // from a stale height nor be restarted by content resizes or duration changes mid-collapse
+    const cssDuration = parseFloat(container?.style.getPropertyValue('--duration') || '') || 0;
```
aleksei-berezkin, line 87: - maybe `getComputedStyle()`?
- I [added](https://github.com/JetBrains/ring-ui/blob/ddafe0e06b73c6594f4533caafe73b3963569385/src/global/parse-css-duration.ts) a util in 8.0, please consider adding a TODO for merging into 8.0

aleksei-berezkin (approved):

andrey-skl (approved):

src/collapse/collapse-content.tsx
Hypnosphi: accessing `style` property directly should be fine as we're setting `--duration` ourselves with inline styles: https://github.com/JetBrains/ring-ui/blob/f1889e4c19a7d6f99a5e0022b434b136af30b2bb/src/collapse/collapse-content.tsx#L89

PR #9380, opened 2026-08-21
RG-2809 Add Trusted Types policy hook

andrey-skl:
## Motivation
Ring UI injects SVG and rich content through `innerHTML` in a few shared components. Host applications that enable [Trusted Types](https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API) can reject those string assignments, so Ring UI needs a single way to receive and use the host's Trusted Types policy instead of forcing every consumer to work around each component separately.

Issue: [RG-2809](https://youtrack.jetbrains.com/issue/RG-2809)

## Summary
- Add a global Trusted Types policy hook to Ring UI configuration
- Apply the configured policy to Icon SVG and ContentEditable innerHTML sinks
- Cover both consumers with regression tests

### DIFF
--- a/src/global/configuration.ts
+++ b/src/global/configuration.ts
@@ -4,14 +4,20 @@ export enum ControlsHeight {
   L = 'L',
 }
 
+export interface RingUITrustedTypePolicy {
+  createHTML(html: string): TrustedHTML;
+}
+
 export interface RingUIConfiguration {
   controlsHeight?: ControlsHeight;
   popupsCssPositioning?: boolean;
+  trustedTypePolicy?: RingUITrustedTypePolicy | null;
 }
 
 const globalConfiguration: Required<RingUIConfiguration> = {
   controlsHeight: ControlsHeight.M,
   popupsCssPositioning: true,
+  trustedTypePolicy: null,
 };
 
 export function configure(config: RingUIConfiguration): void {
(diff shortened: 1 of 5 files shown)


src/global/configuration.ts
```diff
@@ -4,14 +4,20 @@ export enum ControlsHeight {
   L = 'L',
 }
 
+export interface RingUITrustedTypesPolicy {
+  createHTML(html: string): string | TrustedHTML;
+}
```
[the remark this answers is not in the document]
andrey-skl: Applied

src/global/configuration.ts
```diff
@@ -4,14 +4,20 @@ export enum ControlsHeight {
...
 }
 
+export interface RingUITrustedTypesPolicy {
+  createHTML(html: string): TrustedHTML;
+}
+
 export interface RingUIConfiguration {
   controlsHeight?: ControlsHeight;
   popupsCssPositioning?: boolean;
+  trustedTypesPolicy?: RingUITrustedTypesPolicy | null;
```
vkrol: Let's name the type `RingUITrustedTypePolicy` (`Types` -> `Type`) and the field name `trustedTypePolicy` for consistency with the canonical name https://developer.mozilla.org/en-US/docs/Web/API/TrustedTypePolicy.

vkrol (approved):

src/global/configuration.ts
andrey-skl: Applied, thanks!

PR #9388, opened 2026-08-24
fix(tooltip): TW-98285 anchor block triggers correctly

Hypnosphi:
[TW-98285](https://youtrack.jetbrains.com/issue/TW-98285) "Expand" tooltip blocks the expand sidebar button

This only happens with js positioning when a block element (e.g. div) is wrapped in a tooltip. The fix allows users to pass a different TagName for correct display

### DIFF
--- a/src/tooltip/tooltip.tsx
+++ b/src/tooltip/tooltip.tsx
@@ -19,7 +19,8 @@ interface Context {
 
 const TooltipContext = createContext<Context | undefined>(undefined);
 
-export interface TooltipProps extends Omit<AllHTMLAttributes<HTMLSpanElement>, 'title'> {
+export interface TooltipProps extends Omit<AllHTMLAttributes<HTMLElement>, 'title'> {
+  TagName?: 'span' | 'div' | null | undefined;
   delay?: number | null | undefined;
   selfOverflowOnly?: boolean | null | undefined;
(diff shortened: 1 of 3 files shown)


src/tooltip/tooltip.tsx
```diff
@@ -96,24 +99,43 @@ export default class Tooltip extends Component<TooltipProps> {
   };
 
   showPopup = () => {
+    const {containerNode} = this;
```
andrey-skl: So it is a generic approach – tooltip will always try to show away from it's child, correct?

andrey-skl (approved):

src/tooltip/tooltip.tsx
Hypnosphi: I just realised that it won't work properly in a case of multiple children, will convert it to draft for now

src/tooltip/tooltip.tsx
```diff
@@ -190,6 +192,7 @@ export default class Tooltip extends Component<TooltipProps> {
       typeof title === 'string' && Tooltip.isShow({title, hide}) ? {'aria-label': title, role: 'tooltip'} : {};
 
     const {onNestedTooltipShow, onNestedTooltipHide} = this;
+    const WrapperElement = (TagName || 'span') as 'span';
```
Hypnosphi: I'm gonna change the props type instead

